PR #11, opened 2020-02-11
Fix "set pipefail" syntax

marcelm:
`set -e pipefail` will not work as intended since `pipefail` must be set with `-o pipefail`. No error occurs at the moment because `set` assigns all the extra arguments to $1, $2, etc.

### DIFF
--- a/src/setup-conda.ts
+++ b/src/setup-conda.ts
@@ -426,7 +426,7 @@ conda activate ${activateEnvironment}`;
   let bashExtraText: string = `
 # ----------------------------------------------------------------------------
 # Conda Setup Action: Basic configuration
-set -e pipefail`;
+set -eo pipefail`;
   if (isValidActivate) {
     bashExtraText += `
 # Conda Setup Action: Custom activation


goanpeca:
Thanks!

So what was happening?

marcelm:
The intention of `set -eo pipefail` is to let things like `non-existing-command | cat` fail. With `set -e pipefail`, it does not fail.

I just saw the command in the log output. I didn’t experience this problem, but it could lead to hard-to-find bugs (CI passes although there’s an error).

goanpeca:
>The intention of set -eo pipefail is to let things like non-existing-command | cat fail. With set -e pipefail, it does not fail.

>I just saw the command in the log output. I didn’t experience this problem, but it could lead to hard-to-find bugs (CI passes although there’s an error).

Indeed! thanks, will make a release in a couple of days and try to add 32 bit support and fix an issue on windows.

marcelm:
Thanks :-)

goanpeca:
Thank you :-) !

Issue #4, opened 2019-12-20
Expose configuration options as inputs


goanpeca:
Some of these options could e exposed directly as inputs to make life easy on subsequent commands

https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#general-configuration

Like: 
- `channels`
- `default_channels`

...

sodre:
@goanpeca, 
Thanks for working on this. It might be useful to also expose the ability to load a .github/.condarc file if it exists. Then the order can be left up to you, either the file loads first and the options in the action yaml override those, or the other way around.

goanpeca:
There is an option to load a file aleeady. See the Readme

sodre:
I missed it the first time I saw it! Thank you so much!

dcdenu4:
It would be great to directly specify the `channels` and other inputs as to avoid carrying around a condarc file in the repo.

goanpeca:
Indeed! I added the config file since that was the broader solution and then Incould start adding more options :-).

In a work or so I will make a new release

goanpeca:
@dcdenu4 I do have some questions here. Channels has the option of appending and prepending, so should both options be exposed or using channels should prepend (give higher priority always above defaults)

This is what I feel would work best, but wanted to hear some comments.

dcdenu4:
@goanpeca I like the prepend approach of giving what's listed priority, in the order it's listed. That is what would be desirable for our use case. The reason we are specifying our own channels is because we want those used over the defaults first.

goanpeca:
Hi @dcdenu4. Just made the release. Let me know how it goes.

PR #17, opened 2020-03-19
PR: Add conda channels option

goanpeca:
Related to #4


goanpeca:
@dcdenu4 could you test this PR ?

You can use this branch as `goanpeca/setup-miniconda@enh/channels`

dcdenu4:
@goanpeca I tested it on my simple use case and it appears to work properly and pull from the added channels. It looks like for `channels: conda-forge, anaconda` , the priority order is `anaconda`, then `conda-forge`. It's a little counter intuitive to me, as I would think the first thing I list has the greatest priority, but if it's documented well then it doesn't really matter. 

Thanks for this!

goanpeca:
That order is weird indeed I think I have an issue there so that is not right.

I will need to look i to it then

goanpeca:
Ok I knew that could be an issue https://zellwk.com/blog/async-await-in-loops/

I need to change to a normal for otherwise the order is not guaranteed.  

And I learned something new about Js :-p

goanpeca:
@dcdenu4 fixed the issue now. It should be in the right order now!

dcdenu4:
@goanpeca Just ran two tests and they both use the first channel in my list, `conda-forge` in my case. So this is expected and different than prior when it was using `anacaonda`, my 2nd specified channel. Looks good!

goanpeca:
Thanks for testing @dcdenu4 making a release soon ;-)

PR #22, opened 2020-04-14
Use repo.anaconda.com for download

jjhelmus:
Use repo.anaconda.com rather than repo.contiuum.io to download Miniconda.  The later is deprecated while the former is preferred.


goanpeca:
Hi @jjhelmus looks good thanks :-)

Let me fix the failing "tests" and I will get back to you

Issue #42, opened 2020-05-24
Supporting arbitrary constructors?


bollwyvl:
Thanks for this tool! 

While the name of the project says it all, it seems like the action could install and use any constructor-generated installer: the immediate candidate would be [miniforge](https://github.com/conda-forge/miniforge/releases), and I'm kinda itching to try out their pypy releases for speedier CI (haven't really benchmarked it yet). But I also see the potential for just any-old-constructor, especially as part of a longer pipeline.

The most blunt instrument, just an `installer-url`, could override all the detection, listing, etc. mechanisms. I'm very new to github actions, but would be interested in looking into this, if you are amenable!

bollwyvl:
> The most blunt instrument, just an installer-url

Something along the lines of:
```patch
diff --git a/src/setup.ts b/src/setup.ts
index a6f4b2b..1bab5b8 100644
--- a/src/setup.ts
+++ b/src/setup.ts
@@ -212,6 +212,29 @@ async function downloadMiniconda(
   return { ok: true, data: downloadPath };
 }
 
+async function downloadInstaller(url: string): Promise<Result> {
+  let downloadPath: string;
+
+  // Look for cache to use
+  const cachedInstallerPath = tc.find(url, url);
+
+  if (cachedInstallerPath) {
+    core.info(`Found cache at ${cachedInstallerPath}`);
+    downloadPath = cachedInstallerPath;
+  } else {
+    try {
+      downloadPath = await tc.downloadTool(url);
+      const options = { recursive: true, force: false };
+
+      core.info(`Saving cache...`);
+      await tc.cacheFile(downloadPath, url, url, url);
+    } catch (err) {
+      return { ok: false, error: err };
+    }
+  }
+  return { ok: true, data: downloadPath };
+}
+
 /**
  * Install Miniconda
  *
@@ -542,6 +565,7 @@ async function applyCondaConfiguration(
  * Main conda setup method to handle all configuration options
  */
 async function setupMiniconda(
+  installerUrl: string,
   minicondaVersion: string,
   architecture: string,
   condaVersion: string,
@@ -571,7 +595,13 @@ async function setupMiniconda(
       };
     }
 
-    if (minicondaVersion !== "" || architecture !== "x64") {
+    if (installerUrl !== "") {
+      utils.consoleLog("\n# Downloading Custom Installer...\n");
+      useBundled = false;
+      result = await downloadInstaller(installerUrl);
[...]

PR #43, opened 2020-05-24
add custom installer url config option

bollwyvl:
Here's the proposed approach for #42.

I didn't commit with all the built changes, as it seems harder to review... please let me know if you want those committed...

There are a few other things it might be nice to add that i ran into while working this up, but maybe not on this PR:

- the top-level function signature is long, and i made it longer :crying_cat_face: 
  - perhaps move the config options into an interface, e.g. `IOptions`
  - it would then also be possible to generate markdown docs, or even linter schema from it
    - or go the other way, and store the schema, and generate the typings from that
- `ncc` is very cool!
  - requiring `npm i -g ncc` is kinda weird... it seems like having that locked as a `devDependency` might be more robust
- formatting after building seems weird
  - formatting the markdown would be good
- actual linting with `eslint` would be good
- in the hooks, `git commit .` is pretty heavy, wasn't expecting that (had a `_NOTES.md` i was writing, didn't need that!)
- in contributing, `npm run pack` isn't a thing and fails


bollwyvl:
Oh: i see that to actually run the tests, I'd have to commit my changes to the build. welp, here we go!

bollwyvl:
Ha, it didn't run: I guess i don't understand how it works!

goanpeca:
Thanks for working on this @bollwyvl 

>the top-level function signature is long, and i made it longer 😿
Yep 🤷 or we could make it not have a top level function and use run directly?

>perhaps move the config options into an interface, e.g. IOptions
Yes that could be nice

>it would then also be possible to generate markdown docs, or even linter schema from it
or go the other way, and store the schema, and generate the typings from that

I have no idea about that, this was my first TS project :-p 

>ncc is very cool!

It is :-p 

>requiring npm i -g ncc is kinda weird... it seems like having that locked as a devDependency might be more robust

I think I used those instructions from github, but if you advice for it, sure :-) 

>formatting after building seems weird

True :-) 

>formatting the markdown would be good

Easy piecy

>actual linting with eslint would be good

🤷 

>in the hooks, git commit . is pretty heavy, wasn't expecting that (had a _NOTES.md i was writing, didn't need that!)

It is, it just made my life easy :-) 

>in contributing, npm run pack isn't a thing and fails

That is correct, I forgot to erase it.

I will take a look at why it is failing :-)

bollwyvl:
[Ooh, :grinning:](https://github.com/goanpeca/setup-miniconda/pull/43/checks?check_run_id=703913886#step:4:28) :
```
user-agent : conda/4.8.3 requests/2.23.0 PyPy/3.6.9 Linux/5.3.0-1020-azure ubuntu/18.04.4 glibc/2.27
```

goanpeca:
Seems to be working now :-)

goanpeca:
Looks good to me @bollwyvl awesome addition :-), I can merge, clean up the contributing and make a 1.5 release.

Anything else you want to add to this? I am happy with how it is. I will take a look at the other suggestions you made.

bollwyvl:
I replaced my hacked-in branch in the example, let it fail, and review away!

goanpeca:
>I replaced my hacked-in branch in the example, let it fail, and review away!

Yes I actually need to add some tests for this (any help in setting that up is welcome!), that do not require changing the action to the current branch, I actually tried using the workflow context syntax, but it seems github expects only text and not expressions.

.github/workflows/example-5.yml
```diff
@@ -0,0 +1,29 @@
+name: "Example 5: Custom installer"
+
+on:
+  pull_request:
+    branches:
+    - '*'
+  push:
+    branches:
+    - '1.x'
```
goanpeca, line 9: I removed the 1.x branch so just master should be fine.

.github/workflows/example-5.yml
```diff
@@ -0,0 +1,29 @@
...
+    - 'master'
+
+jobs:
+  example-5:
+    name: Ex5 Miniforge for PyPy
+    runs-on: 'ubuntu-latest'
```
goanpeca, line 15: Will this work on other platforms? if so, maybe we should also add them

bollwyvl: Pypy would work on `osx-64`, but not on `win-64`

goanpeca (changes requested):
Couple of minor changes, otherwise looks good!

.github/workflows/example-5.yml
goanpeca: Maybe then we should add a note somewhere? README 

And make this `runs-on: ['ubuntu-latest', 'macos....]`

bollwyvl: added in 723a49a

bollwyvl:
Yeah, so osx ran, but didn't actually try it, and didn't fail (just warned)... i'll run it once with my branch...

.github/workflows/example-5.yml
bollwyvl: worked like a boss:
```
             user-agent : conda/4.8.3 requests/2.23.0 PyPy/3.6.9 Darwin/19.4.0 OSX/10.15.4
```

will put it back

.github/workflows/example-5.yml
bollwyvl: are you wanting a change here from me? seems like something that should be done to all of them in one fell swoop...

.github/workflows/example-5.yml
bollwyvl: it would be hard to `matrix` these: `constructor` by default generates predictable filenames, but they can be overridden (or renamed with, e.g. a build number), so i'm more inclined to leave it verbose at this point...

goanpeca: Verbose is fine for this

goanpeca:
Thanks for this @bollwyvl 

Keep testing with `goanpeca/setup-miniconda@master` while I make the release and let me know how it goes.

PR #44, opened 2020-05-24
Add eslint and ncc to dev deps

bollwyvl:
Here's a quick pass at adding and getting `eslint` to be happy. Also set up prettier to go after all the files it can, because why not.

Adding it to this repo's workflow as a fast fail should be pretty straightforward, and will save some coal burning for easily-fixable errors.

Probably won't have time to continue messing about with it today, but might already be an interesting review... i yanked most of the defaults from jupyterlab/jupyterlab, but there's room for interpretation!


bollwyvl:
It appears that cross-workflow job dependencies aren't possible, or require some fairly significant investment. I guess at this point, it's quick enough to not require _too_ much optimization.

src/setup.ts
```diff
@@ -608,17 +621,18 @@ async function setupMiniconda(
       utils.consoleLog("\n# Downloading Custom Installer...\n");
       useBundled = false;
       result = await downloadInstaller(installerUrl);
+      if (!result.ok) return result;
```
bollwyvl, line 624: amid all the other changes, this is actually something i missed from #43

.github/workflows/caching-example.yml
```diff
@@ -23,7 +22,7 @@ jobs:
         with:
           path: ~/conda_pkgs_dir
           key: ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('etc/example-environment.yml') }}
-      - uses: goanpeca/setup-miniconda@master
+      - uses: ./
```
goanpeca, line 25: Nice :-p 

!

goanpeca:
>It appears that cross-workflow job dependencies aren't possible, or require some fairly significant investment. I guess at this point, it's quick enough to not require too much optimization.

Yeah, among jobs we could but not among workflows, unless we used artifact uploading, so not worth it.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,21 @@
...
+    branches:
+    - 'master'
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+      - shell: bash -l {0}
```
goanpeca, line 17: Should we use a specific node version and action? Cause we are now relying on what comes in the container, which could change.

Thougths?

.github/workflows/caching-example.yml
bollwyvl: yeah, that was a good one to find...

.github/workflows/lint.yml
bollwyvl: As an action, we kinda get `node-12` and that's that, right?

src/setup.ts
goanpeca: Got it, I missed it too!

goanpeca (commented):
Awesome addition @bollwyvl !

Left one question, otherwise LGTM!

Thanks :-)

.github/workflows/lint.yml
bollwyvl: Anyhow, I'm open to pinning... suggestions on what the keys/values should be?

goanpeca: In theory :-p 

Let's leae like that for now, merging

goanpeca:
Will make a 1.5 release tomorrow!

Issue #14, opened 2020-03-04
Add mamba support


nh13:
I have found that using mamba for dependency solving complex environments MUCH faster.  You [previously mentioned](https://github.com/goanpeca/setup-miniconda/issues/13#issuecomment-594088885) you may be convinced to add support.   I am making a separate issue to track this.

goanpeca:
Sure @nh13 I will add a flag just to see how things work, however there is the issue of using conda-forge which is where mamba lives. So I need to check if I need to also change the default channel to point to conda-forge instead of anaconda defaults.

jaimergp:
I'd be tremendously interested in this! Let me know if I can help somehow.

goanpeca:
Hi @jaimergp 

Thanks for the interest.

It is just a matter of adding the option for it and installing it. :-)

I guess

mamba-version. To have control over what version to use and then replace the conda command with mamba.

Want to tackle it? PRs welcome

jaimergp:
Thanks for your answer!

What do you think of this approach using Minimamba? https://quantstack.net/mamba.html

PR #47, opened 2020-06-09
add mamba support

jaimergp:
As discussed per #14.

This is my first TS GitHub Action, so I might have done things in a weird way. Also, I don't know how to test locally, so let's see if it works. 

Please let me know how I can improve the TS idioms too!

### DIFF
--- a/.github/workflows/example-6.yml
+++ b/.github/workflows/example-6.yml
@@ -1,0 +1,33 @@
+name: "Example 6: Mamba"
+
+on:
+  pull_request:
+    branches:
+    - '*'
+  push:
+    branches:
+    - 'master'
+
+jobs:
+  example-6:
+    name: Ex6 Mamba
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./
+        with:
+          python-version: 3.6
+          mamba-version: latest
+          channels: conda-forge,defaults
+          channel-priority: true
+          activate-environment: anaconda-client-env
+          environment-file: etc/example-environment.yml
+      - shell: bash -l {0}
+        run: |
+          conda info
+          conda list
+          conda config --show-sources
+          conda config --show
+          printenv | sort
+      - shell: bash -l {0}
+        run: mamba install jupyterlab
--- a/.github/workflows/example-6.yml
+++ b/.github/workflows/example-6.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: ./
         with:
           python-version: 3.6
-          mamba-version: latest
+          mamba-version: "*"
           channels: conda-forge,defaults
           channel-priority: true
           activate-environment: anaconda-client-env
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 ![Example 3: Other options](https://github.com/goanpeca/setup-miniconda/workflows/Example%203:%20Other%20options/badge.svg?branch=master)
 ![Example 4: Channels](https://github.com/goanpeca/setup-miniconda/workflows/Example%204:%20Channels/badge.svg?branch=master)
 ![Example 5: Custom installer](https://github.com/goanpeca/setup-miniconda/workflows/Example%205:%20Custom%20installer/badge.svg?branch=master)
+![Example 6: Mamba](https://github.com/goanpeca/setup-miniconda/workflows/Example%206:%20Mamba/badge.svg?branch=master)
 ![Caching Example](https://github.com/goanpeca/setup-miniconda/workflows/Caching%20Example/badge.svg?branch=master)
 ![Linting](https://github.com/goanpeca/setup-miniconda/workflows/Linting/badge.svg?branch=master)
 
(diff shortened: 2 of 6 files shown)


jaimergp:
Welp, seems to work! Unfortunately, there's an issue in how environments are installed in this action.

Mamba [does not support `mamba env update` yet](https://github.com/QuantStack/mamba/issues/42). In this action, the environment file is treated in a different way than the usual `conda env create -f <file.yml>`. Instead, it first creates an empty environment with only `python <python-version>` and then it updates it with the YML. Is there any reason behind this alternative strategy? Looks a bit redundant because you need two solving steps instead of a single one.

goanpeca:
@jaimergp thanks for the PR. The reason was to simplify the logic but then id mamba has that limitation then we should probably change it.

jaimergp:
> @jaimergp thanks for the PR. The reason was to simplify the logic but then id mamba has that limitation then we should probably change it.

Seems that they might ["look into it" today](https://gitter.im/QuantStack/Lobby?at=5edf7252ef5c1c28f0106903). Let's see :)

jaimergp:
In the meantime, I have given it some thought (and this is probably out of the scope of this PR, so feel free to branch the discussion into a separate issue).

Right now, there are several ways a new environment gets initialized with this action:

* `base` gets initialized by default, optionally changing the default python version.  This can be the builtin Miniconda or an externally provided one.
* A custom env gets created if an environment file is created (or is a custom name is passed?). The way this works is by creating an empty environment first, then installing Python, then updating that with the environment definitions. However, this might have (mostly harmless) side effects, such as having Python in the environment even if the user didn't specify that in the YML file. It also leads to longer wait times because there are two (or three, depending on .condarc) solving steps: (empty env), Python installation, environment update.

I think environment creation from a file should be decoupled from the Custom env + Python of your choosing setup. It's less problematic, faster and the code shouldn't be much more complicated.

What do you think? This could be enforced with errors and warnings too. For example if both python-version and environment-file are specified, it should be an error IMO, since the Python version of your choice should be specified in the YML anyway. For matrix builds, this is solved with multiple env files if needed or other conda magics.

goanpeca:
>I think environment creation from a file should be decoupled from the Custom env + Python of your choosing setup. It's less problematic, faster and the code shouldn't be much more complicated.

Yes, and do

>What do you think? This could be enforced with errors and warnings too. For example if both python-version and environment-file are specified, it should be an error IMO, since the Python version of your choice should be specified in the YML anyway. For matrix builds, this is solved with multiple env files if needed or other conda magics.

This is precisely the reason why it is split into steps, you could have a single yaml file with python specified, but not the actual version. A lot of projects need to test with different versions and doing this way is what allows for that.

On these points

>base gets initialized by default, optionally changing the default python version. This can be the builtin Miniconda or an externally provided one.

I don't understand what you mean by this.

>However, this might have (mostly harmless) side effects, such as having Python in the environment even if the user didn't specify that in the YML file.

Not exactly true, as explained above is python is a requirement but several versions will be tested the python in the env yaml should have no version.

>It also leads to longer wait times because there are two (or three, depending on .condarc) solving steps: (empty env), Python installation, environment update.

Not really. An empty env has no solver, so the waiting time is practically zero. A python installation takes no time to solve since it will be placed on an empty env and uses well-established dependencies. environment update-> In here the Python version will be used as a constraint for the new environment and that is the reason why python version will not be changed at all.

It is not possible to have a single env create step and be able to provide a python version, since the command that runs that is condo env not conda, so something like:

`conda end create -f someenv.yml python=$PYTHON_VERSION` will not work.

And yes this is a separate issue and a separate thread and exposed before,  the only viable change for current workflows (that use the possibility of changing the python version without hardcoding it on the env.yaml) is to reduce going from 3 steps to just two so that creating the empty env and adding python to it is a single step. In reality, the gain in that is like 1 second, so not really worth the trouble.

jaimergp:
Thanks for the detailed answer!

Is it possible then to have an environment set up without Python? Let's say, we have this file:

```yaml
name: test
channels:
- conda-forge
- defaults
packages:
- cmake
- clang-dev
```

goanpeca:
>Is it possible then to have an environment set up without Python? Let's say, we have this file:

Yes, if `python-version`  is not set or if it is not on the yaml file, and it is not a dependency of the listed requirements, there is no reason why python would get installed (unless there is a bug :-p )

See: https://github.com/goanpeca/setup-miniconda/blob/master/src/setup.ts#L759

jaimergp:
Awesome! Thanks for the explanation again, then my main concerns were based on misconceptions :) Apologies for the time taken 😬

goanpeca:
Hmmm good points

Since mamba is a package and condo does not have the notion of latest (miniconda installers do) I would be inclined in having to provide an actual version number. So I am still not 100% on either of the suggestions, since that would hold for python-version: "latest"  for instance.

Mamba is an "advanced" feature so you should be explicit about the version you want if you are already on that path.

What do you think @bollwyvl ?

jaimergp:
You are right, `conda` happily takes `*` as a version specifier. No need to use default or auto or anything.

jaimergp:
Btw, `mamba` devs fixed this! https://github.com/QuantStack/mamba/pull/327 They are also using your GH Action ;)

goanpeca:
Haha awesome and great to hear

wolfv:
I have just released mamba 0.3.2 
Once this is green https://github.com/conda-forge/mamba-feedstock/pull/48 you could give it another try :)

goanpeca:
Awesome! Thanks @wolfv !

goanpeca:
@jaimergp is everything working as expected? Do you need to make an update to the examples to check for the fix?

jaimergp:
Looks like the new mamba is indeed applied with `mamba env update`!

* [Mamba 0.3.1](https://github.com/goanpeca/setup-miniconda/runs/754488616#step:3:568)
* [Mamba 0.3.2](https://github.com/goanpeca/setup-miniconda/pull/47/checks?check_run_id=755567901#step:3:567)

I am a bit confused because of how the downloads look, though. In a regular `mamba install` they look [like this](https://github.com/goanpeca/setup-miniconda/pull/47/checks?check_run_id=755567901#step:3:554), but with `mamba env update` they look like the [regular conda downloads](https://github.com/goanpeca/setup-miniconda/pull/47/checks?check_run_id=755567901#step:3:633). In a nutshell, looks like solving is done with mamba but not downloading. Is this intentional @wolfv?

wolfv:
HI @jaimergp I think you're right. With `mamba env` the downloads are still done through conda. I can fix that

jaimergp:
Thank you so much!

wolfv:
hey, mamba 0.3.3 (it's in the process of being released to conda-forge) has an important fix for an issue that manifested itself in CI environments sometimes. 

Just wanted to make sure that you're not stuck on 0.3.2!

Cheers & thanks for your work here!

goanpeca:
Thanks again @jaimergp and @wolfv :-)

goanpeca:
Ok version 1.6 is out 🥳 !

jaimergp:
Yayy thanks everybody!!

PR #52, opened 2020-06-30
Add architecture option to installation candidates

### LINKED ISSUE #10: Allow 32 bit option
https://github.com/goanpeca/setup-miniconda/blob/bde5ac1a373f18066038802241b6bf0ccf747d8a/src/main.ts#L25

Currently, it always defaults to 64 bit. It would be nice to use this to build and test conda 32 bit packages on Windows.

bryanwweber:
The user can select an architecture supported by their Actions runner as
an option in the with block.

Closes #10

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -82,6 +82,10 @@ inputs:
     description: 'Experimental. Use mamba (https://github.com/QuantStack/mamba) as a faster drop-in replacement for conda installs. Disabled by default. To enable, use "*" or a "x.y" version string.'
     required: false
     default: ""
+  architecture:
+    description: 'Architecture of Miniconda that should be installed. Available options on GitHub-hosted runners are "x86" and "x64". Default is "x64".'
+    required: false
+    default: "x64"
 runs:
   using: "node12"
   main: "dist/setup/index.js"
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -21823,10 +21823,16 @@ function setupMiniconda(installerUrl, minicondaVersion, architecture, condaVersi
                 result = yield installMiniconda(result.data, useBundled);
             }
             else if (minicondaVersion !== "" || architecture !== "x64") {
-                if (architecture !== "x64" && minicondaVersion == "") {
+                if (architecture !== "x64" && minicondaVersion === "") {
                     return {
                         ok: false,
                         error: new Error(`"architecture" is set to something other than "x64" so "miniconda-version" must be set as well.`)
+                    };
+                }
+                if (architecture === "x86" && process.platform === "linux") {
+                    return {
+                        ok: false,
+                        error: new Error(`32-bit Linux is not supported by recent versions of Miniconda`)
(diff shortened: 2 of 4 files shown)


bryanwweber:
This appears not to have done anything so far, see: https://github.com/Cantera/conda-recipes/pull/22/checks?check_run_id=823649924#step:3:15 which is still using the pre-installed Miniconda

goanpeca:
Hi @bryanwweber thanks for taking look. Now, we need to force a miniconda install from scratch for this to work, since it is not possible to "change" the architecture and user the bundled one. So if `miniconda-version` is not provided, we need to use "latest" to force a install.

bryanwweber:
Thanks for the feedback! I had interpreted this line:
https://github.com/goanpeca/setup-miniconda/blob/03cf5633af2fd0e98baa281a7f2803c2417d7c83/src/setup.ts#L654

as "if minicondaVersion is set OR architecture isn't equal to x64", in which case, when I set `architecture` to `x86`, it should go through the install steps. Did I miss something?

Looking at it now, I see that `minicondaVersion` might have to be set in addition to `architecture`. Or rather, if `architecture` is set, then `minicondaVersion` must also be set. Should I add a conditional that checks for that?

joergbrech:
@bryanwweber, @goanpeca, I really like the setup-miniconda action and I could use the architecture option for several projects! Great work! What ist the status on this PR?

BTW, I played around with this the `bryanwweber:fix-10` branch, and it works great, except for one issue with x86 Miniconda on ubuntu-latest:

https://github.com/DLR-SC/tigl-conda/runs/965599369?check_suite_focus=true#step:3:66

Do you have an idea what this is about? Is this a permission issue?

bryanwweber:
@joergbrech It seems as though Anaconda have dropped support for x86 Linux, the installer is from January of last year (2019) 
https://repo.anaconda.com/miniconda/ I would not suggest relying on that working. You may want to try and set up a local x86 Linux machine to debug, and if you find something, I'd be happy to include a fix here.

goanpeca:
Hi @bryanwweber, sorry I have swamped with work, and I still have to review this!

---

Thanks @joergbrech, indeed I do not think the Linux issue is solvable (we should probably disallow it @bryanwweber so )

joergbrech:
> It seems as though Anaconda have dropped support for x86 Linux, the installer is from January of last year (2019)

Ah, I did not know that. Thanks for the info.  I am looking forward to this being merged! Thanks again for the great work!

bryanwweber:
Thanks @goanpeca I'll try to add another check for that condition. No problem about the delay, we ended up just dropping Win32 support for our package rather than trying to build the conda package, so I ended up not needing this feature 😄

bryanwweber:
@goanpeca I've pushed a change that errors out when 32-bit Linux is used

goanpeca:
Awesome! I will merge an iterate! Thanks!

PR #55, opened 2020-07-15
Support custom_channels

xhochy:
Currently there is no way to support custom channel configurations, e.g. channels served over https with username/password not on anaconda.org.

### DIFF
--- a/src/setup.ts
+++ b/src/setup.ts
@@ -925,6 +925,23 @@ async function run(): Promise<void> {
       show_channel_urls: showChannelUrls,
       use_only_tar_bz2: useOnlyTarBz2
     };
+
+    // Extract custom channels from the configuration.
+    if (channels.length != 0) {
+      utils.consoleLog("Splitting " + channels);
+      let split_channels: Array<string> = channels.split(",");
+      let channel: string;
+      for (channel of split_channels) {
+        let key: string = "custom_channels." + channel;
+        utils.consoleLog("Checking for " + key);
+        let custom_channel: string = core.getInput(key);
+        if (custom_channel.length != 0) {
+          utils.consoleLog("Setting " + key);
+          condaConfig[key] = custom_channel;
+        }
+      }
+    }
+
     const result = await setupMiniconda(
       installerUrl,
       minicondaVersion,
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -22030,6 +22030,21 @@ function run() {
                 show_channel_urls: showChannelUrls,
                 use_only_tar_bz2: useOnlyTarBz2
             };
+            // Extract custom channels from the configuration.
+            if (channels.length != 0) {
+                utils.consoleLog("Splitting " + channels);
+                let split_channels = channels.split(",");
+                let channel;
+                for (channel of split_channels) {
+                    let key = "custom_channels." + channel;
+                    utils.consoleLog("Checking for " + key);
+                    let custom_channel = core.getInput(key);
+                    if (custom_channel.length != 0) {
+                        utils.consoleLog("Setting " + key);
+                        condaConfig[key] = custom_channel;
+                    }
+                }
+            }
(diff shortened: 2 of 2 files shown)


goanpeca:
Hi @xhochy, thanks for the contribution.

I do not understand what you are trying to accomplish. Could you add an example, and more information for context?

Also, how can this be tested?

Thanks!

xhochy:
I have a custom conda channel hosted on a private http server. Thus I cannot simply add the channel via the name but also need to provide the URL explicitly via `custom_channels`. One caveat is that you can only supply a `Dict[str, str]` under the `with:` key in the Github Actions YAML. Thus we have to flatten the key names here.

```
channels: my_channel,conda-forge
custom_channels.my_channel: https://${{ secrets.CONDA_CHANNEL_UPLOAD_USER }}:${{ secrets.CONDA_CHANNEL_UPLOAD_PASSWORD }}@conda-host.example.com
```

goanpeca:
@xhochy got it now thanks! 
>One caveat is that you can only supply a Dict[str, str] under the with: key in the Github Actions YAML. Thus we have to flatten the key names here.

Yeah, that is a pain. This is a convenience to be able to use things from the yaml file directly, right? Using the condard option would have still allowed you to do this with the current version. Correct?

bollwyvl:
Sorry we've been letting this sleep! Hadn't actually looked over this one.

I wonder: the conda config space is _huge_, and as we've learned, changes over time, and tracking every single feature might be challenging. Might it be easier to just add a `condarc` field as a string?  YAML-in-YAML is gross, but these things don't get _that much_ validation anyway

```yaml

    - uses: conda-incubator/setup-miniconda@v1
        with:
          condarc-file: etc/example-condarc.yml
          condarc: |
            custom_channels:
              my_channel: https://${{ secrets.CONDA_CHANNEL_UPLOAD_USER }}:${{ secrets.CONDA_CHANNEL_UPLOAD_PASSWORD }}@conda-host.example.com
```

It could follow something like:
- if `condarc-file` is specified, copy it into place
  - otherwise, initialize an empty one
- if `condarc` is specified, merge that in
- read channels, etc. off the final file

just spitballing...

goanpeca:
Hi @xhochy, sorry this has been standing for so long... you know... time... ⏳ 

>I wonder: the conda config space is huge, and as we've learned, changes over time, and tracking every single feature might be challenging. Might it be easier to just add a condarc field as a string? YAML-in-YAML is gross, but these things don't get that much validation anyway

Indeed it is huge, but maybe we could just expose all of them, plus the dot notation, `custom_channels.my_channel:` that actually does not sound so bad.

Yes more work on parsing on our side, but we can use conda itself to do all the validation for us. Just need to define the type of expected types and the `dot` parser.

>YAML-in-YAML is gross, 

Indeed and I would rather support the inputs, than an error prone string syntax.

src/setup.ts
```diff
@@ -925,6 +925,23 @@ async function run(): Promise<void> {
...
       use_only_tar_bz2: useOnlyTarBz2
     };
+
+    // Extract custom channels from the configuration.
+    if (channels.length != 0) {
+      utils.consoleLog("Splitting " + channels);
+      let split_channels: Array<string> = channels.split(",");
+      let channel: string;
+      for (channel of split_channels) {
+        let key: string = "custom_channels." + channel;
```
goanpeca, line 935: I would rather this kept the dash, instead of the underscore, for the sake of consistency with the other input params.

goanpeca:
Actually now that I think of it, we cannot use this as the input parameter would become variable and input params needs to be fixed and defined in here:

https://github.com/conda-incubator/setup-miniconda/blob/master/action.yml#L4

As it is standing the only option I would suggest is to use the `condarc-file` input and have envirnoment variables in the channels to be replaced for the user and password?

goanpeca:
Since this is not possible I think this can be closed, but we should track this on a new/separate issue.

PR #61, opened 2020-08-14
Allow explicit conda environment specs

mathieu1:
Here's a first go at supporting [explicit spec files](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#building-identical-conda-environments). The main upshot is that it bypasses completely the solving step (which can speed up the action significantly) and produces a fixed environment (that can only break if the packages become unavailable).

It's working, but there are currently no safegards nor documentation. There are definitely a few caveats to using explicit conda specs: 
- It usually doesn't make sense to specify a Python version (it's usually already in the spec)
- One must provide an `activate-environment` name (it's not in the spec)
- One must take care to provide a spec for the right platform

I haven't tested what would happen with a bundled miniconda, what happens when using mamba, or how it interacts with caching.

Is there interest in further work on this?

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -21870,12 +21870,14 @@ function setupMiniconda(installerUrl, minicondaVersion, architecture, condaVersi
             }
             // Read the environment yaml to use channels if provided and avoid conda solver conflicts
             let environmentYaml;
+            let environmentExplicit;
             if (environmentFile) {
                 try {
                     const sourceEnvironmentPath = path.join(process.env["GITHUB_WORKSPACE"] || "", environmentFile);
-                    if (fs
+                    environmentExplicit = fs
                         .readFileSync(sourceEnvironmentPath, "utf8")
-                        .includes("\n@EXPLICIT\n")) {
+                        .includes("\n@EXPLICIT\n");
+                    if (environmentExplicit) {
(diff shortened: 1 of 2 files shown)


bollwyvl:
Thanks for doing this. it will work _really well_ with [conda-lock](https://github.com/mariusvniekerk/conda-lock).

goanpeca:
Thanks a lot for working o this. Please go ahead!

Any additional input Nick?

bollwyvl:
No specific feedback,  nor do i have a handy example, as I haven't evaluated using conda-lock + mamba + setup-miniconda... because all these things _weren't_ available not too long ago.

On this PR, I guess it would take:
- a reasonable example
- a minimal test
- docs

To get it over the line. 

Interesting, of course, would be:
- an absolute-craziest case, which would pile on
  - pypy miniforge
  - mamba
  - caching
  - a checked-in explicit file

I don't think any special handling of conda-lock would be _required_ here, as a developer would use it before committing to generate the `@EXPLICIT` files, which would then be referenced here. However, I could imagine that being a future feature, where the generated lock files get cached, too.

goanpeca:
Hi @mathieu1 sorry for the delay. Could you rebase or merge with master so we can merge :-) ?

PR #67, opened 2020-08-27
Fix README

HiromuHota:


goanpeca:
Hi @HiromuHota thanks for the fix!

PR #68, opened 2020-09-03
Update readme for caching

### LINKED ISSUE #65: Caching troubles & questions
I'm trying to take advantage of caching of dependencies. It seems like whatever I do the`setup-miniconda` action takes just as long regardless of whether I've cached my dependencies or not.

```yaml
    - name: Cache conda
      uses: actions/cache@v1
      env:
        # Increase this value to reset cache if environment.yml has not changed
        CACHE_NUMBER: 5
      with:
        path: ~/conda_pkgs_dir
        key: ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('environment.yml') }}
    - uses: goanpeca/setup-miniconda@v1
      with:
        activate-environment: myenvname
        channel-priority: strict
        environment-file: environment.yml
        use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
```
Have I made a mistake here that is preventing `setup-miniconda` from loading the packages from `~/conda_pkgs_dir` as specified in the docs? 

**Important note** Possibly worth documenting, removing `python-version: 3.7` from the step resulted in a faster build, but still no advantage when caching. I was seeing much slower build times as the step would hang on installing python 3.7.

My environment file looks a bit like this:
```yaml
name: myenvname
channels:
  - pytorch
  - anaconda
  - conda-forge
  - defaults
dependencies:
  - python=3.7
  - jupyter=1.0.0
  - numpy=1.17.4
  - pandas=0.25.3
  - pytorch=1.4.0
  - pip=20.0.2
  - pip:
    - coverage==5.0.3
```

It seems like some dependencies are downloading regardless of the fact that they were stored in the cache (the caching step is successful and it shows a reasonable size).
```zsh
# Updating conda environment from yaml file...
##############################################

/usr/share/miniconda/condabin/conda env update --file environment.yml --name myenvname
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... done
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
Ran pip subprocess with arguments:
['/usr/share/miniconda/envs/my-org/bin/python', '-m', 'pip', 'install', '-U', '-r', '/home/runner/work/my-org/my-repo/condaenv.p1divyve.requirements.txt']
Pip subprocess output:
Collecting coverage==5.0.3
  Downloading coverage-5.0.3-cp37-cp37m-manylinux1_x86_64.whl (227 kB)
```

It seems like it's not picking up the files that were restored from the cache, what am I missing?
Also should the updated `cache@v2` be supported?

MGough:
Applied suggestions to README.md regarding caching pip packages separately. As per #65 

I've also moved the declaration of the python version for the caching example from inside the workflow, to inside `etc/example-environment.yml`. I don't think that this change should affect the other workflows that use this file, hopefully the github actions will pick up on this. I would carry out further testing but unfortunately I'm about to lose internet access for about a week! Feel free to modify this if needed.

Closes #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,8 @@ jobs:
           use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
 ```
 
+If you are using pip to resolve any dependencies in your conda environment then you may want to [cache those dependencies separately](https://docs.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#caching-dependencies), as they are not included in the conda package cache.
+
 ## IMPORTANT
 
 - Bash shells do not use `~/.profile` or `~/.bashrc` so these shells need to be explicitely declared as `shell: bash -l {0}` on steps that need to be properly activated. This is because bash shells are executed with `bash --noprofile --norc -eo pipefail {0}` thus ignoring updated on bash profile files made by `conda init bash`. See [Github Actions Documentation](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#using-a-specific-shell) and [thread](https://github.community/t5/GitHub-Actions/How-to-share-shell-profile-between-steps-or-how-to-use-nvm-rvm/td-p/33185).
--- a/README.md
+++ b/README.md
@@ -273,7 +273,6 @@ jobs:
       - uses: goanpeca/setup-miniconda@v1
         with:
           activate-environment: anaconda-client-env
-          python-version: 3.8
           channel-priority: strict
           environment-file: etc/example-environment.yml
           use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ jobs:
         with:
           activate-environment: anaconda-client-env
           channel-priority: strict
-          environment-file: etc/example-environment.yml
+          environment-file: etc/example-environment-caching.yml
           use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
 ```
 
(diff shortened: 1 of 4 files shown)


MGough:
Looking at the Mamba run as an example, it seems that as I'd suspected the declaration in the `environment.yml` overrides the one in the workflow. Perhaps a separate `environment.yml` could be used for the caching example?

goanpeca:
>Looking at the Mamba run as an example, it seems that as I'd suspected the python version declaration in the environment.yml overrides the one in the workflow. Perhaps a separate environment.yml could be used for the caching example?

Agreed, please make one :-) 

And thanks a lot for the PR!

README.md
```diff
@@ -273,12 +273,13 @@ jobs:
       - uses: goanpeca/setup-miniconda@v1
```
MGough, line 273: I've just noticed that this is referring to your old repo  (@goanpeca ) before this was absorbed into `conda-incubator`. I'd update these myself but I'm afraid I have very limited time with internet before I'll be cut off for a while, and I wouldn't want to miss any. A simple but very minor issue could be raised from this.

MGough:
I've swapped it over to use a new environment file. Unfortunately I won't be able to make any more changes for around a week, but hopefully it's all good now. I have spotted another minor issue in the the action name being used in the repo, it's unrelated to my changes.

README.md
goanpeca: I can fix those, no worries 👍

goanpeca:
>'ve swapped it over to use a new environment file. Unfortunately I won't be able to make any more changes for around a week, but hopefully it's all good now. I have spotted another minor issue in the the action name being used in the repo, it's unrelated to my changes.

It is good to go, I will fix the old urls.

Thanks a lot for helping with this! I really appreciate the help @MGough

Issue #70, opened 2020-09-14
New repo URL breaks existing workflows


jaimergp:
I would have expected runners to follow redirections for actions that have changed location, but they don't...

https://github.com/openforcefield/openforcefield/runs/1114396909?check_suite_focus=true

Thanks to @mattwthompson for the report!

goanpeca:
Hi @jaimergp, yes I was expecting the redirects to last longer :-|

mattwthompson:
It does seem to be working as expected after manually updating it: https://github.com/openforcefield/openforcefield/actions/runs/254560374

goanpeca:
Thanks for the report @mattwthompson, leaving open as this will probably bite more people :-|

jaimergp:
Are all users expected to manually migrate though? Maybe you can post the last version to your `goanpeca` namespace too, but adding a warning that it's changing locations and that the repo won't receive any more updates?

goanpeca:
>Are all users expected to manually migrate though? Maybe you can post the last version to your goanpeca namespace too, but adding a warning that it's changing locations and that the repo won't receive any more updates?

Yep that would have been a better path 😆 

I do not have time to do this at the moment so I will just create a fork and add a warning later when time clears a bit.

Thanks for the suggestion.

goanpeca:
Created this fork, so I guess this should bring it back to life ?

https://github.com/goanpeca/setup-miniconda

jaimergp:
Thank you so much for the fast response! That will do in the meantime, yes!

jaimergp:
I also raised this through their Twitter support and they are looking into it:

https://twitter.com/jaime_rgp/status/1305620040595763204

jaimergp:
Twitter support has theoretically addressed this: https://twitter.com/mscccc/status/1306002367175823360

I don't know what we should do now, though:

a) Remove the fork and hope the redirection works
b) Keep the fork but add a deprecation warning pointing to the new URL

goanpeca:
Let's go with b :-)

That would be the nice thing to do.

Now for how long would redirects work?

jaimergp:
> I would favor b :-)

Agreed!

> Now for how long would redirects work?

As far as I know, they shouldn't expire? [source](https://twitter.com/mscccc/status/1308780155142156288)

goanpeca:
Let's keep the fork for now while we add the new things.

bollwyvl:
[...]

PR #71, opened 2020-09-16
update URLs in README.md from goanpeca to conda-incubator

MGough:
This was breaking the badges in the readme. I've also updated the workflow examples, which hopefully will reduce new users accidentally encountering problems related to #70 

Previously discussed in #68


goanpeca:
Thanks a lot @MGough! 

Looks good :-), merging.

PR #74, opened 2020-09-19
Extend #61 with tests and docs

bollwyvl:
A follow-on to #61, this starts tests and documentation for the explicit spec approach. It could be merged back into #61, but I didn't want to start pushing to @mathieu1's `master`...


bollwyvl:
This one's looking good... 25s vs 35s isn't much, for this solve, but on windows, with a long environment and a heavy dep like `scikit-image`, not doing the solve provides a really nice speed-up.

bollwyvl:
:tada:

PR #80, opened 2020-10-06
Test all platforms with explicit example

bollwyvl:
Looks like I failed to add a windows test for #74, and windows line endings are not picked up by the `\n@EXPLICIT\n` check, leading to brokenedness. Hopefully the regex `^` will figure it out...


bollwyvl:
The old `conda` on win-64 cannot deal with `.conda` files...  presumably there is a way to make sure conda-lock doesn't do that. in the meatime, replaced the example solve with a lock of `conda-lock` itself, which doesn't generate publish `.conda` packages, only `.tar.bz2`... yet... sigh.

bollwyvl:
Great, that worked. Realized that the `channels` warning is also meaningless in the case of an explicit, so have squashed that as well...

goanpeca (approved):
Looks good 👍 

Thanks again @bollwyvl

goanpeca:
So maybe now is time to make a new release?

bollwyvl:
My guinea pig with the lock files at least got to that point:

https://github.com/krassowski/jupyterlab-lsp/pull/348/checks?check_run_id=1216668382

...so :+1: to release from my side

PR #82, opened 2020-10-08
Add a link from each CI badge to the corresponding section in the README

mwouts:
Hi @goanpeca , I was glad to see that you've been working on `mamba` support!

What would you think of adding these links? With this, clicking on a badge (other than 'Linting') will get the user to the corresponding section in the README.

PR #86, opened 2020-10-08
Table with links to sections and CI builds

mwouts:
This is an alternative to #82. In this version the TOC appears as a table, see https://github.com/mwouts/setup-miniconda/blob/toc_as_table/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,41 @@
 # Setup Miniconda
 
-![Example 1: Basic usage](https://github.com/conda-incubator/setup-miniconda/workflows/Example%201:%20Basic%20usage/badge.svg?branch=master)
-![Example 2: Other shells](https://github.com/conda-incubator/setup-miniconda/workflows/Example%202:%20Other%20shells/badge.svg?branch=master)
-![Example 3: Other options](https://github.com/conda-incubator/setup-miniconda/workflows/Example%203:%20Other%20options/badge.svg?branch=master)
-![Example 4: Channels](https://github.com/conda-incubator/setup-miniconda/workflows/Example%204:%20Channels/badge.svg?branch=master)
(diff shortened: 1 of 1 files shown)


bollwyvl:
Which one do you like better, now that you're looking at them?

mwouts:
I think I prefer #82, as the badges are more 'compact', see https://github.com/mwouts/setup-miniconda/blob/master/README.md
Don't you think that the table is a bit big? Maybe it's better if the reader does not need to scroll to read the first paragraph.
Anyway that's just my preference, feel free to choose whichever you like most!

bollwyvl:
It is tall, and will get taller... while it might fit today as a wide table, scrolling horizontally is an even worse user experience, for sure. My issue with the super-long-words-in-a-picture is it is very small text... and the `Example X: ` prefix adds almost no information, but I don't think we should change the names at this point. 

Let's close this, unless you want to try anything else, and concentrate on #82...

goanpeca:
Hello @mwouts, thanks for working on this, and sorry for the delays.

> In this version the TOC appears as a table, see

I think I actually like this version better, as indeed it works as a TOC and is easier to read vertically, than horizontally with badges of varying size. It does take more vertical space, but I think it helps with guiding users to an specific section in the README.


Thoughts?

mwouts:
Hi @goanpeca , well the table version is longer for sure, but you're right, it makes it more explicit that the left entries are links to the sections.

Personally I'm good with any version that lets the user go to the matching paragraph! This means either
- #82 (but we lose the link to the CI)
- This one, #86 
- Or even a simple table of contents separated from the CI statuses. You can generate the TOC simply by copy pasting the README at https://github.com/luciopaiva/markdown-toc . Maybe that form of TOC would be simpler to maintain?

bollwyvl:
We adapted and updated this over on #153, thanks again and sorry for the delay. Feel free to open a new issue if the current setup isn't too your liking!!

PR #94, opened 2020-11-16
Try to catch EnvironmentSectionNotValid

jaimergp:
Fixes #93

PR #95, opened 2020-11-16
Fix CI for example 8

jaimergp:
CI is working, but the code behind #94 is not!

### DIFF
--- a/.github/workflows/example-8.yml
+++ b/.github/workflows/example-8.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           auto-update-conda: true
           python-version: ${{ matrix.python-version }}
-          environment-file: etc/example-faulty-environment.yml
+          environment-file: etc/example-environment.yml
       - name: Check previous step failed
         shell: bash
         if: always()
--- a/.github/workflows/example-8.yml
+++ b/.github/workflows/example-8.yml
@@ -27,6 +27,7 @@ jobs:
           environment-file: etc/example-faulty-environment.yml
       - name: Check previous step failed
         shell: bash
+        if: always()
         run: ${{ steps.setup-miniconda.conclusion == 'failure' }}
       - name: Conda info
         shell: bash -l {0}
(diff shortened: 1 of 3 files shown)


jaimergp:
I don't know what's wrong. The listeners do not seem to intercept the streams at all :/

goanpeca:
@jaimergp I think you need to do the build step so that the assets get compiled and then that is what is used by actions. 

Are you running
```
npm run format
npm run check
npm run build
```

jaimergp:
🤦  I thought that was being done on the precommit hook because it took a bit longer than usual. Heh 😬

goanpeca:
It works :) !

.github/workflows/example-8.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           environment-file: etc/example-faulty-environment.yml
       - name: Check previous step failed
         shell: bash
-        run: ${{ steps.setup-miniconda.conclusion == "failure" }}
+        run: ${{ steps.setup-miniconda.conclusion == 'failure' }}
```
goanpeca, line 30: So since the previous step fails, this will never be reached.

goanpeca: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

jaimergp: Hm, I fixed that and it runs now but the CI result is still red. I thought last successful one counted but maybe we have to setup that explicitly?

jaimergp:
Ok, finally found the right combo. To check for expected failures:

* The step that is expected to fail must have `continue-on-error: true` and an `id`.
* The evaluation step needs `if: always()` and it should check for `outcome == 'failure'` not `conclusion == 'failure'`.

This is ready now!

goanpeca:
>Ok, finally found the right combo. To check for expected failures:

Awesome work @jaimergp. Thanks a lot!

PR #98, opened 2020-11-16
Refactor tool downloading and output, add Windows MiniForge example

### LINKED ISSUE #97: installer-url option triggers an error on windows
I'm trying to use the miniforge distribution, it works just as expected for linux and osx, but on windows it fails when trying to cache the downloaded installer, and it logs the following output:

```
# 
# Downloading Custom Installer...

#####################################

Miniforge3-Windows-x86_64.exe
Saving cache...
Error: EINVAL: invalid argument, mkdir 'C:\hostedtoolcache\windows\https:\github.com\conda-forge\miniforge\releases\latest\download\Miniforge3-Windows-x86_64.exe\https:\github.com\conda-forge\miniforge\releases\latest\download\Miniforge3-Windows-x86_64.exe\x64'
```

I suspect the error could be due to the additional `:` characters in the directory path, coming from the url schema?

Thanks a lot for developing this action,

Best,
Riccardo

bollwyvl:
- fixes #97

Code Changes:
- rework the download to unify the behavior of the miniconda downloader (which was obvioiusly working) and the custom one
- rework outputs
- in example 5, update miniforge, add windows examples, use "nasty" urls with `&?#`


src/setup.ts
```diff
@@ -293,8 +293,9 @@ async function downloadInstaller(url: string): Promise<Result> {
     downloadPath = cachedInstallerPath;
   } else {
     try {
+      core.info(`Downloading to...\n\t${cachedInstallerPath}`);
       downloadPath = await tc.downloadTool(url);
-      core.info(`Saving cache...`);
+      core.info(`Saving to cahce...\n\t${downloadPath}`);
```
goanpeca, line 298: Typo 🙃

bollwyvl: yah, gotcha. i might try out the `::fold` thing...

bollwyvl:
Very close now...

.github/workflows/example-5.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./
         with:
-          installer-url: https://github.com/conda-forge/miniforge/releases/download/4.8.3-2/Miniforge-pypy3-4.8.3-2-MacOSX-x86_64.sh
+          installer-url: https://github.com/conda-forge/miniforge/releases/download/4.9.0-3/Miniforge-pypy3-4.9.0-3-MacOSX-x86_64.sh?foo=bar&baz
```
goanpeca, line 42: `?foo=bar&baz` Is this related to cache?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## [v2.1.0][] (Unreleased)
+
+## Fixes
+
+- [#97][] fixes `installer-url` on Windows
```
goanpeca, line 7: Hmm good point should have added this to the other PRs :-p 

@jaimergp would you like to update this on the changelog with the PRs you started?

goanpeca (approved):
Awesome work @bollwyvl 

Thanks a lot!

.github/workflows/example-5.yml
goanpeca: Oh nevermind, you used it to test the url parsing.

Right @bollwyvl ?

bollwyvl:
Thanks for the review! 

Looking over the logs, `tc.find` never seems to get a cache hit... will investigate.

bollwyvl:
Ah, I was banging my head against paths and escaping: turns out `tar` is just broken on windows: https://github.com/actions/toolkit/issues/632

I still think we want to use the `exec(cmd[0], cmd.slice(1))` everywhere, just seems like fewer surprises...

.github/workflows/example-5.yml
bollwyvl: Yep, just want to make sure we've got an out for the inevitable s3 trash or wherever people are putting these things. Actually, going to add some docs about that...

goanpeca:
>I still think we want to use the exec(cmd[0], cmd.slice(1)) everywhere, just seems like fewer surprises...

Agreed. Good to merge then?

bollwyvl:
Sorry my attention is a bit async right now... it looks like `file:///` works like a boss (probably a PITA on windows, but so it goes), but this is pretty clutch for being able to test an installer built somewhere else during a workflow without turning on a web server or something.

No further changes planned!

goanpeca:
>No further changes planned!

You pinky promise?

bollwyvl:
we swears. I'm working conda-forge chores now :tired_face:

CHANGELOG.md
jaimergp: Oops, yep, will do!

goanpeca: No worries I forgot too 🤦🏽

PR #100, opened 2020-11-17
add tests for mamba on windows and osx

jaimergp:
Trying to investigate and (hopefully) solve #59

PR #102, opened 2020-11-17
add changelog entries for #94 #95 #100

jaimergp:
Address https://github.com/conda-incubator/setup-miniconda/pull/98#discussion_r525379796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@
 
 ## Fixes
 
-- [#97][] fixes `installer-url` on Windows
+- [#97][] fixes `installer-url` on Windows.
 - [#94][], [#95][] catches ignored errors when an environment file contains
-  invalid section names [#93][]
+  invalid section names [#93][].
 - [#100][] fixes `mamba` not being available on Windows if using a `bash` shell
-  [#59][]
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 ## Fixes
 
 - [#97][] fixes `installer-url` on Windows
+- [#94][], [#95][] catch ignored errors when an environment file contains
+  invalid section names
+- [#100][] fix `mamba` not being available on Windows if using a `bash` shell
```
bollwyvl, line 10: Ooops, i missed my link target, as do these... just needs the refs like the one below... they are all `pulls` except #97, which is under `issues. I guess as to convention:
- it's talking about "(the release) fixes" or "(the release) catches"... not a big deal, but makes it more consistent to read
- for fixes, if there is an original issue, that number is probably more interesting, as (as we've seen) it can take a couple PRs to finally fix something

jaimergp: Heh, ok. I thought somehow that link targets were added at the end. Would we be interested in having a PR template with all these things?

goanpeca:
Thanks for working on this @jaimergp :) !

Issue #75, opened 2020-09-23
Micromamba based setup?


jaimergp:
Thoughts on this? https://github.com/mamba-org/mamba#micromamba

> micromamba is a tiny version of the mamba package manager. It is a pure C++ package with a separate command line interface. It can be used to bootstrap environments (as an alternative to miniconda), but it's currently experimental. The benefit is that it's very tiny and does not come with a default version of Python.

It might be too soon because it's still experimental, but it could be helpful in the future.

goanpeca:
Hi @jaimergp 

See the discussion on #66

mariusvniekerk:
So ensureconda (https://github.com/conda-incubator/ensureconda) is a tiny little tool that like pip ensures that a conda (or conda-like) is around.  

It has some built in ways to install the standalone conda and micromamba.  Generally not intended as a replacement for things like miniconda but more for conda related tools like `conda-lock`.  Might suit some of your needs?

mariusvniekerk:
We can certainly integrate some of the features from ensureconda into setup-miniconda if it makes sense

goanpeca:
Hi @mariusvniekerk thanks for it. I have been looking at ensure conda for some work I will be doing in constructor, so thanks for that. 

The ideal place would be to have micromamba bundled with the github runners. Otherwise downloading things would be almost the same for the bundled miniconda (installing mamba) then downloading micromamba.

But to have things added to github runners (besides enough requests) is project maturity I think ?

mariusvniekerk:
Yeah main difference is mostly the micromamba standalone installation doesn't need conda at all, so you save a conda solve for mamba.

hadim:
The "Setup Conda" step when used with `mamba` seems to take a long time (~6 min). A `mamba` only setup would probably fix this.

goanpeca:
>The "Setup Conda" step when used with mamba seems to take a long time without installing my custom deps only mamba-version to "*" (~6 min). A mamba only setup would probably fix this.

Yep that is possible. Investigating these delays and trying to mitigate them.

bollwyvl:
Provided #126 goes through, this is ready for business, as in https://github.com/conda-incubator/setup-miniconda/issues/76#issuecomment-748717482

The difference here is that `useMamba` would need to get hoisted, earlier, and more forcefully, as no `conda` command would be available. It's also probably _not_ a `constructor`-compatible CLI, but I could be wrong... if not, we'll need to do some updating in a few more places.

[...]

PR #107, opened 2020-11-21
[do not merge] major refactor of downloads, environment creation

bollwyvl:
As a prelude to #75/#76, we talked about splitting out the downloader pieces into smaller things.

This is that... and entirely too much else. I basically couldn't track where all the `if` statements were going, and this is what I came up with to make future providers of installers more tractable... micromamba wouldn't look like what's in there, but we'll get there when we get there, I guess.

- move _most_ of the validation to parse time to fail earlier
  - more can be done here 
- strategy pattern (one per very short file, except miniconda download) applied to: 
  - installers sources
  - base tools (e.g. mamba, conda-build)
  - environment creation 
  - baseline packages (e.g. python, though now can add more)
- does at most one solve, seems quite snappy
- more consistent top-level logging approach
- standardized _most_ functions to take... 
  - a frozen `inputs` (as from the yaml, this could still be improved
  - a dynamic grab-bag of things (e.g. `useMamba`, the parsed env file)
    - some _more_ typing could be applied here
- normalize package spec string creation
- removed most of the `any` types
- removed the `(Error|Success)Result` pattern, leaning almost entirely on `throw` and one top-level `catch` (except around settings)

Anyhow... I would worry about this bitrotting pretty fast, as any other features/fixes change, but it's also going to be a _bear_ to review. So sorry! I can _try_ to pull it apart into smaller incremental pieces, but basically did the whole thing in one go when my network was patchy.

## Sub-PR Plan

- [x] #108 get to this basic file structure (no subdirectories) without changing any code so subsequent prs are targeted at individual files, and not just looking at the "god" function
  - by-the-by: will probably set `.gitattributes` for `dist` to be binary... it's never worth reviewing
- [x] #109 refactor `Result` to native `throw`
- [x] #123 refactor the input parsing, 
- [x] #125 use inputs throughout (along with the mutable state object)
- [x] #126 refactor the installer provider
- [x] #127 refactor the env provider
- [ ] refactor the tool provider


goanpeca:
Hi @bollwyvl as always thanks for this. This is a bit too much for a single PR so if we could split this, that would be great. Also less chance of breaking things.

bollwyvl:
> Question what is the deal with underscored files?

Just for sorting. could be changed.

>  too much for a single PR

Sure, fair. Will add a top-level plan of action...

PR #108, opened 2020-11-22
[1 of #107] Split up setup.ts

bollwyvl:
The `setup.ts` file is getting unwieldy. To make future PRs more easily reviewed, this PR will break it up into some smaller files.

- first part of #107
- [x] marks all js as binary, as nobody reviews them anyway
- [x] runs linter against cron stuff
- [x] rebuilds because :shrug: 
- [x] break up existing constants, functions, and types into smaller files
  -  won't touch `setupMiniconda`


goanpeca:
@bollwyvl could you rebase. Then we should be good to go and iterate.

goanpeca:
Any comments @conda-incubator/setup-miniconda-team ? :)

PR #109, opened 2020-11-22
[2 of #107] Use native Error

bollwyvl:
The `Result` pattern adds a lot of line noise for the extra `try` and `catch`, and is rather inflexible in its return types. This PR removes _most_ of the `try/catch` blocks, the `*Result` types, and all references to them.

- continues #107 
- includes:
  - #108 
    - recommend viewing only changes from https://github.com/conda-incubator/setup-miniconda/pull/109/commits/0e849fe)... 
    - or better yet, `git diff -w` locally, as a lot of the `setup.ts` diff is unreadable in the web UI

### DIFF
--- a/dist/delete/index.js
+++ b/dist/delete/index.js
@@ -1102,13 +1102,10 @@ function execute(command) {
                 },
             },
         };
-        try {
-            yield exec.exec(command[0], command.slice(1), options);
+        const rc = yield exec.exec(command[0], command.slice(1), options);
+        if (rc !== 0) {
+            throw new Error(`${command[0]} return error code ${rc}`);
         }
-        catch (err) {
-            return { ok: false, error: err };
-        }
-        return { ok: true, data: "ok" };
     });
 }
 exports.execute = execute;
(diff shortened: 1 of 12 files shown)


goanpeca:
@bollwyvl I think all the same changes from the original big PR ended up here?

Having changes that mix changes and also split things into files make me very nervous and are harder to review.

So my suggestion for this work. Split things first, or change code first but not both, that is too complex.

I would go for let's first change the file structure without changing anythin else, not a single line of code, just copy pasting from one place to another, that way we are sure things are not breaking in subtle and unexpected ways.

From there we can go PR by PR to change a single thing (like this PR attempted)

Cheers @bollwyvl and thanks again for all the hard work!

bollwyvl:
This PR only offers 0e849fe _on top of_ #108, which does the file split.

goanpeca:
Duh. Silly me. Thanks for clarifying. :)

goanpeca:
@bollwyvl maybe do a merge/rebase with master?

bollwyvl:
Yep, I'll get on it presently, and fire up the next couple steps.

bollwyvl:
Let's wait to just squash merge, rather than re-write what's in this PR...

bollwyvl:
Looking forward to more PRs with changes like `+525 −812`, before we're through...

bollwyvl:
Great, thanks. Will keep keeping on in subsequent prs...

PR #124, opened 2020-12-20
[4 of #107] replace destructuring with input member access

bollwyvl:
This finishes the work in #123 by removing the huge object destructuring in `setup`, and uses `inputs.<action input>` throughout. It also removes the configuration checks which will have already been handled at _parsing_ time by `inputs`. This allows for dropping a few imports :tada:.

- continues #107 

## Follow-ons:

This further reveals `condaConfig`, and the rest of the "mutable state" during the run is down to:

```ts
  const condaConfig = { ...inputs.condaConfig };
  let useBundled: boolean = true;
  let useMamba: boolean = false;
```

though additional mutable variables get added further down. However, the only part of the `condaConfig` that gets changed is channels, and actually gets co-mingled with what is read from an environment. Looks like #120 bears on this as well.

### Proposal

From [107](https://github.com/conda-incubator/setup-miniconda/pull/107/files#diff-3067e4ddc9cec53d1e1b16104c7ab358b069bbf5754b96ac825f3274fc42c04bR78)

```ts
**
 * Options that may change during the course of discovery/installation/configuration
 */
export interface IDynamicOptions {
  useBundled: boolean;
  useMamba?: boolean;
  envSpec?: IEnvSpec;
}
```

There might be a better name, but it's a little easier to demonstrate when the surrounding code is brought over: it replaces most function specs with `(inputs, dynamicOptions)`. Things get a little messy with `mamba`, etc. but that lies further ahead...

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -21425,7 +21425,6 @@ var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, ge
 Object.defineProperty(exports, "__esModule", { value: true });
 const fs = __importStar(__webpack_require__(747));
 const path = __importStar(__webpack_require__(622));
-const url_1 = __webpack_require__(835);
 const core = __importStar(__webpack_require__(470));
 const io = __importStar(__webpack_require__(1));
 const yaml = __importStar(__webpack_require__(414));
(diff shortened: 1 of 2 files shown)


goanpeca:
Hi @bollwyvl thanks for working on this.  `IDynamicOptions` sounds good. Do you want to add that to this PR or a separate one?

Cheers!

bollwyvl:
Unless you'd prefer to see the larger story, I feel like `+91 −165` is a good place to stop on this one (as it was mostly stuff not done in part 3, due to those _other_ changes). The next one will start to touch a lot more files, but generally in a good way.

goanpeca:
Let's iterate on this then :)

PR #125, opened 2020-12-20
[wip] [5 of #107] add dynamic options

bollwyvl:
This encapsulates the state that _does_ change, as highlighted in #124, in a single, relatively lenient object, and changes most function signatures to accept inputs, options or both.

- continues #107 

> @bollwyvl: I haven't tested this one out. :crossed_fingers: 

## On the next PR

> refactor the installer provider

One of the first "irregularities" is that `runInstaller` is allowed to calculate its own `outputPath`, which requires access to mutable state. If we split up that behavior, we can get a lot of reuse, where it will be more straightforward to add additional "well-known" installer providers.

The idea here is to break up the installer provisioning step even more. There should be a significant net negative line count in `setup.ts`, and encapsulate the logic more succinctly along with its effects.

This is where we want to leave enough breathing room to handle micromamba #75 without actually _building_ it right now.

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -5920,9 +5920,9 @@ exports.ensureLocalInstaller = ensureLocalInstaller;
  *
  * @param installerPath must have an appropriate extension for this platform
  */
-function runInstaller(installerPath, useBundled) {
+function runInstaller(installerPath, options) {
     return __awaiter(this, void 0, void 0, function* () {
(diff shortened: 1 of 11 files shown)


goanpeca:
LGTM @bollwyvl gave it a quick look :)

bollwyvl:
Yeah, had another look I don't see anything else glaring at this step.

PR #126, opened 2020-12-21
[6 of #107] move installer detection logic to implementation

bollwyvl:
This moves all of the boolean logic for each installer provider out of `setup` to its individual file. This 1) cuts down on the complexity in `setup`, and 2) makes it very tractable to add the next installer, e.g. MiniForge, etc.

- continues #107

## On the next PR

> refactor the env provider

Apply a similar treatment as this PR to the _running_ of an installer and creating the environment. Another large chunk of `setup` will drop, and the boolean switches will be closer to their specific implementations. This is where the hook would become relevant for whether `mamba` is already provided by the installer for, e.g. config operations.

### DIFF
--- a/src/installer/bundled-miniconda.ts
+++ b/src/installer/bundled-miniconda.ts
@@ -1,0 +1,29 @@
+import * as types from "../types";
+
+/**
+ * Provide a path to the pre-bundled (but probably old) Miniconda base installation
+ *
+ * ### Note
+ * This is the "cheapest" provider: if miniconda is already on disk, it can be
+ * fastest to avoid the download/install and use what's already on the image.
+ */
+export const bundledMinicondaUser: types.IInstallerProvider = {
+  label: "use bundled Miniconda",
+  provides: async (inputs, options) => {
+    return (
+      inputs.minicondaVersion === "" &&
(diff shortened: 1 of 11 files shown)


src/installer/bundled-miniconda.ts
```diff
@@ -0,0 +1,29 @@
...
+  label: "use bundled Miniconda",
+  provides: async (inputs, options) => {
+    return (
+      inputs.minicondaVersion === "" &&
+      inputs.architecture === "x64" &&
+      inputs.installerUrl === ""
+    );
+  },
+  installerPath: async (inputs, options) => {
+    // no-op
```
goanpeca, line 20: What does this mean @bollwyvl ?

goanpeca (approved):
LGTM, just one question, but not blocking the merge.

Thanks @bollwyvl 🥳 !

src/installer/bundled-miniconda.ts
bollwyvl: so with the bundled installer, we do _nothing_, as it's already installed... just update the options.

goanpeca: Got it.

Issue #122, opened 2020-12-19
Override environment.yaml name


JulioV:
Is there a way to override the name of the environment created from an `environment.yaml` file (a.k.a. the `--name` flag)?

BTW, thanks for your work, it saved me a lot fo time!

goanpeca:
Hi @JulioV, thanks for your suggestion and your kind words :)

>Is there a way to override the name of the environment created from an environment.yaml file (a.k.a. the --name flag)?

Well could you describe a bit more what you are trying to accomplish?

JulioV:
Our conda environment is named like this `projectYYYYMM`, so every time the `name` attributes changes in our `environment.yaml` file I need to change `activate-environment` in your action. Is there a way to pass an environment name to your action, so that it does not have to match the `name` attribute in the file?

goanpeca:
>Is there a way to pass an environment name to your action, so that it does not have to match the name attribute in the file?

`test-environment` serves this purpose, but I think they need to be the same, or it needs to be missing from he env yaml, so I don't think we support you use case at the moment.

JulioV:
Oh OK, can I transform this issue into a future request? :)

jaimergp:
You could patch the `environment.yml` file in a previous step with `sed` and friends. Ugly but would work for now :)

JulioV:
@jaimergp, thanks, that's what I thought. For anyone seeing this in the future I did  ` sed -i "" "s/name:.*/name:my_new_name/g" environment.yml`

bollwyvl:
So do you want your environment to get created with the name:
- `test`
- `projectyyymmdd` 
- or some other name specified in your file (other than `base` or `root`)

More broadly: what should the order of preference be (not _your_ preference for this use case, but _least surprising_ for the most people? Practically, I think over on #127 i'm at a place where I can fix this, if it hasn't already done so.

My spitballing hierarchy, from most- to least-explicit:
- no name in env, name in workflow:  whatever was in workflow (if not `base` or `root`)
- name in env, name in workflow: whatever was in workflow (if not `base` or `root`)
- name in env, no name in workflow: ????
  - > i think it's `test` right now, and not sure if anybody likes that, but also don't know if we can change it, now
- no name in env, no name in workflow: `test`

Some thoughts for the future:
- explicit envs already don't have a name, so there's nothing to fall back on
- if multiple env files are supported, what then?

bollwyvl:
[...]

PR #127, opened 2020-12-21
[7 of #107] expand env parsing and command creation 

bollwyvl:
This collapses the boolean logic for parsing an environment file, and adds the parsed environment (as yaml or explicit lines) to the dynamic options. Subsequently, it moves the "create an env" logic to dedicated files for "simple" conda create, create with a lockfile and `env update` with a YAML. This drops a _lot_ from `setup.ts`, and notably a number of transient dependencies (`yaml`, etc).

- continues #107
- includes
  - #126  
- might help #122?

## On the next PR

> refactor the tools provider

Apply a similar treatment as this PR to the provisioning of tools in the base env (e.g. `mamba`, `conda-build`), prior to installation of the environment.

Again, we need a way to say, _hey, my base env already/only has `mamba`_. One Does Not Simply Parse a Conda MatchSpec, though... so I guess I'd rather it was explicit, e.g. `installer-has-mamba`, which we'd be able to pre-populate for `mambaForge` rather than trying to do anything with `mamba-version`.


bollwyvl:
Diff's getting a little nasty, will keep this `wip` until #126 lands...

goanpeca:
#126 landed @bollwyvl

src/types.ts
```diff
@@ -125,3 +125,24 @@ export interface IInstallerProvider {
     options: IDynamicOptions
   ) => Promise<IInstallerResult>;
 }
+
+/**
+ * A strategy for ensure an environment is up-to-date vs the action inputs
```
bollwyvl, line 130: typo, will batch with other review fixes

src/conda.ts
```diff
@@ -62,47 +62,63 @@ export async function bootstrapConfig(): Promise<void> {
...
-          );
-        } catch (err) {
-          core.warning(`Couldn't set conda configuration '${key}'`);
-        }
-      }
-    }
+  // channels are special: if specified as an action input, these take priority
+  let channels = inputs.condaConfig.channels
+    .trim()
+    .split(/[,\n]/)
```
bollwyvl, line 92: this inadvertently allows for `\n` in addition to `,` as a channel separator. having both is probably a bug unless we document it... will revert to existing behavior.

the correct behavior is probably to _only_ use `\n` _if_ there is a (non-trailing?) `\n` in the value

goanpeca: Sounds good. The correct behavior to me would be to have all commas or all line separators.

Could we do something like `/(,)|(\n)/`and count each group?

src/conda.ts
```diff
@@ -62,47 +62,63 @@ export async function bootstrapConfig(): Promise<void> {
   );
 }
 
+/**
+ * Copy the given condarc file into place
+ */
+export async function copyConfig(inputs: types.IActionInputs) {
+  const sourcePath: string = path.join(
+    process.env["GITHUB_WORKSPACE"] || "",
```
goanpeca, line 70: Should we use the home path?

src/conda.ts
```diff
@@ -62,47 +62,63 @@ export async function bootstrapConfig(): Promise<void> {
...
+  }
+
+  // LIFO: reverse order to preserve higher priority as listed in the option
+  for (const channel of channels.reverse()) {
+    core.info(`Adding channel '${channel}'`);
+    await condaCommand(["config", "--add", "channels", channel], options);
   }
 
-  await condaCommand(["config", "--show-sources"], notMambaOptions);
+  // all other options
```
goanpeca, line 106: Feel like all these comments should be capitalized, unless there is some sort of TS convention.

src/constants.ts
```diff
@@ -66,3 +66,21 @@ export const CONDA_CACHE_FOLDER = "conda_pkgs_dir";
...
+ *
+ * ### Note
+ * Some examples:
+ * - python
+ * - python 3
+ * - python>3
+ * - python!=2
+ * - conda-forge::python
+ *
+ * TODO: this should be generalized, and, along with roundtrip parsing/generating
```
goanpeca, line 82: Yeah, could be :)

src/utils.ts
```diff
@@ -48,3 +48,17 @@ export async function execute(command: string[]): Promise<void> {
     throw new Error(`${command[0]} return error code ${rc}`);
   }
 }
+
+/**
+ * create a conda veersion spec string.
```
goanpeca, line 53: Typo

goanpeca (changes requested):
Couple of questions, and also I feel like all comments should be capitalized 🤷🏽 

Thanks for this amazing work @bollwyvl !

src/conda.ts
bollwyvl: I think we need to assume sourcing the config from the workspace, which may or may not be in home.

the `CONDARC_PATH` on the other hand _is_ in `HOME`, as reported by `os.homedir()`

src/conda.ts
bollwyvl: As i said, let's not monkey with the newlines on this PR, will put it back to comma-only.

If, for some awful reason, someone _must_ have a `,` in their URL, as it _is_ part of the RFC, allowing both is going to break, hard. 

Typographically, just commas scans quickly horizontally, while the lines look more like all the other conda docs. I don't see a huge win to supporting both in one, either:
```yaml
channels: |
  conda-forge,defaults
  https://www.crazy.town/conda-remote/
```
one must really parse what each line is saying, and there's no whitespace to help...

goanpeca: Ok, commas for now then :)

bollwyvl:
Hopefully all typos/capitalization handled on
https://github.com/conda-incubator/setup-miniconda/pull/127/commits/64e43746f1144df89ec0c932169bfaec527fe6b3

goanpeca (approved):

Issue #76, opened 2020-09-24
Setup-miniforge


goanpeca:
Hi @conda-incubator/setup-miniconda-team,

I created https://github.com/conda-incubator/setup-miniforge as a convenience for dealing directly with conda forge on defaults.

My plan was to have a script update the parts that needed updating so the it would follow this repo, and just have the script run every now and then to keep the things updated.

Maybe there is a better way of doing this.

Thoughts?

jaimergp:
The changes would be:

* Changing the default Miniconda installer so it points to the Miniforge releases instead
* Hardcoding conda-forge+defaults as the default channel configuration
* Anything else?

If that's the case, I guess it's easy to maintain with a little script, yes. I for one would prefer Miniforge to Miniconda as well.

jaimergp:
Another option would be to deprecate `installer-url` and `miniconda-version` in a single key like `conda-distribution` or something like that. This key would allow _magic keywords_ like "miniconda", "miniforge", or "anaconda", that would default to the latest URL for those installers, but would also accept URLs directly (for example, if users want to use their custom `constructor`-derived installers).

goanpeca:
Hi @jaimergp 

So from your list:

>Changing the default Miniconda installer so it points to the Miniforge releases instead

Sure, but I think we can actually just remove the defult defaults (or update them) when users are not installing a new miniconda/miniforge since. I usually just used the bundled one (which tends to be faster even with updating conda).

>Hardcoding conda-forge+defaults as the default channel configuration

Yep we can do that

>Anything else?

Hmmm make sure the scripts replaces the right things on readmes and maybe just maintain a V1, V2 (major) versions so we do not need to recreate tags.

bollwyvl:
I feel like not breaking back-compat is important, as we have seen. I'd say the v2 line needs to stay with a miniconda default, while v3 would be fine to change the defaults, if not the name. Either way, i'd say it's totally fine to just have prefixes, e.g. `miniforge-*` and `miniconda-*`. Further, miniforge use a slightly different URL scheme (e.g. pypy) than miniconda, so we'd need to handle all their spec bits to make it feel smooth.

The work started on #98 pointed out that we need to beef up the "get me an installer" game, perhaps move it to another folder with a file per strategy, e.g. 
```
download/
  base.ts
  file.ts
  custom.ts
  miniforge.ts
  miniconda.ts
```

as our architectures, etc. will become disjoint.

[...]

PR #133, opened 2020-12-28
Add (Mini/Mamba)forge

bollwyvl:
## Related
- one approach to #76

## Code Changes
- `download-miniforge.ts`
  - uses GitHub API to fetch releases (if version is not specified)... am trying to properly cache these, but i think there's some cache locality stuff
 - removes some dead code
 - tweaks provider logging to avoid constructions like `Can we use use conda create?`
- `conda.ts`
 - if `use-mamba` ~~and `mamba-in-installer`~~ is specified, will try to use `mamba` earlier in the process, but also catches things like `conda init` which _aren't_ supported by the `mamba` CLI yet
- `installer/index.ts`
  - now accepts (and can update) dynamic options, so that it can detect an installed `mamba`  
- `base-tools/index.ts`
  - splits up tool specs and post-install
- `upgrade-mamba.ts`
  - can no-up the upgrading, but we will _still_ want the mamba bash wrapper on windows without starting the solver)
- tests
  - adds workflow with a shotgun of different options to test out versions, interplay with other stuff 

## User-facing changes
- adds `miniforge-variant`, `miniforge-version`, to specify a Miniforge product
- adds ~~`mamba-in-installer`~~ and `use-mamba` to fine-tune use of Mambaforge
- updates the README, with some cross-links between mamba, custom installer, and this new provider

## Backwards-incompatible changes
N/A

## Follow-on

- with these changes, we're ready for `micromamba` (#75), but: 
  - the `mamba` CLI doesn't yet appear to support `init` for shell integration... 
    - perhaps we can hot-install `micromamba` into the bundled miniconda `bin` or `Scripts` or `condabin` or something?
- while our list of inputs (and therefore our README) is getting rather long, I think from a maintenance perspective, keeping _one_ action running that can do all the things is probably useful vs splitting effort (especially now that we've #107-ed ourselves to where we are now)
  - perhaps we should consider doing a real docs site, potentially with a little "wizard" for building up a workflow
  - or, start using the GitHub wiki or discussions

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -11,9 +11,29 @@ inputs:
     default: ""
   miniconda-version:
     description:
-      "If provided, this version of miniconda will be downloaded and installed.
+      "If provided, this version of Miniconda3 will be downloaded and installed.
       Visit https://repo.continuum.io/miniconda/ for more information on
       available versions."
+    required: false
+    default: ""
+  miniforge-variant:
+    description:
+      "If provided, this variant of Miniforge will be downloaded and installed.
+      Currently-known values:
+      - Miniforge3
+      - Miniforge-pypy3
+      - Mambaforge
+      - Mambaforge-pypy3
+      Visit https://github.com/conda-forge/miniforge/releases/ for more information
+      on available variants"
+    required: false
+    default: ""
+  miniforge-version:
+    description:
+      "If provided, this version of the given Miniforge variant will be downloaded
+      and installed instead of the latest.
+      Visit https://github.com/conda-forge/miniforge/releases/ for more information
+      on available versions"
     required: false
     default: ""
   conda-version:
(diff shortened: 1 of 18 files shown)


bollwyvl:
Hooray :green_heart: !

goanpeca:
Thanks for working on this @bollwyvl, this is awesome :)

action.yml
```diff
@@ -180,13 +200,26 @@ inputs:
       removed from the runner. Default is "true".'
     required: false
     default: "true"
+  mamba-in-installer:
```
jaimergp, line 203: Can this be checked programmatically? Once installed one can check for `mamba` or `mamba.bat` in `$PREFIX/condabin` and determine it without user intervention, I think.

bollwyvl: hmmm... it's probably worth a look...

goanpeca:
>perhaps we should consider doing a real docs site, potentially with a little "wizard" for building up a workflow

Absolutely, I think we are big enough that we can have dedicated docs :) will set something with githubpages and myst :)

bollwyvl:
Ok, so ab60b01 is back to :green_circle: and 4651cc0 is mostly cosmetic... thanks @jaimergp for pushing to remove the manual thing: doing the detection uncovered a number of future bugs (e.g. adding that bash wrapper on windows).

`example-10.yml` is getting pretty hairy at this point, but I think has proven out a lot of cases (e.g. mamba in the installer upgrading to a specific mamba). I'd better stop tinkering with it.

Let's _definitely_ squash merge this one :P

jaimergp:
Nice!

>  adding that bash wrapper on windows

Hopefully this is addressed at the mamba level at some point (them providing a shell command like conda does)

bollwyvl:
> Nice!
>
> adding that bash wrapper on windows
>
> Hopefully this is addressed at the mamba level at some point (them
> providing a shell command like conda does)
>
Indeed. Guess we should issue? I'm a little mamba-ed out right now.

God help zsh/fish windows users... But I could see xonsh, frankly...
P-strings are awesome for windows "specialness".

goanpeca:
Another awesome piece of work @bollwyvl :) 🥳 

Thanks!

Issue #136, opened 2020-12-30
Error "The input line is too long" when running vcvarsall.bat in a conda environment


jdumas:
Hi there,

I'm facing a rather annoying issue trying to compile a C++ program in a conda environment. Usually, I would `call vcvarsall.bat` with right args to set up the development environment for Visual Studio. However, since I've updated from `goanpeca/setup-miniconda@v1` to `conda-incubator/setup-miniconda@v2` this no longer works and I encounter the following error:
```
The input line is too long.
The syntax of the command is incorrect.
```

You can see [here](https://github.com/MeshFEM/microstructures/pull/2/checks?check_run_id=1624655242) a run that illustrates the problem. I've tried several things, such as switching to Powershell (but still need to call `vcvarsall.bat` via cmd), or putting the whole line [into quotes](https://stackoverflow.com/questions/682799/what-to-do-with-the-input-line-is-too-long-error-message/3583282#3583282), but these did not work. When I comment out the conda env from the github action then things work again, so I think the combination of conda + vcvarsall.bat is exceeding some limit in an env variable somewhere...

Any idea what might have changes between v1 and v2 that would explain the error now?

goanpeca:
Hi @jdumas thanks for the report!

We will look into it. @bollwyvl any ideas 🙃

goanpeca:
@jdumas could you run your CI and print the content of %PATH% ?

bollwyvl:
yuck, that's one of those files i tend to close my eyes when i see problems related to it.

We do explicitly `core.addPath` to `%CONDA%\condabin` (i didn't put it there, but it was probably for a good reason at the time), and then the shell integration adds a couple more things and `condabin` _again_, so yeah, we're costing you 22 needless characters. But, if indeed it's `%PATH%` length, it's really just "lucky" that _we_ broke it, as it's a trash heap on the hosted runners with something like...

<details>
<summary>this monstrosity...</summary>
<pre>
C:\aliyun-cli
C:\cf-cli
C:\hostedtoolcache\windows\go\1.14.13\x64\bin
C:\hostedtoolcache\windows\Python\3.7.9\x64
C:\hostedtoolcache\windows\Python\3.7.9\x64\Scripts
C:\hostedtoolcache\windows\Ruby\2.5.8\x64\bin
C:\hostedtoolcache\windows\stack\2.5.1\x64
C:\Miniconda\condabin
C:\Miniconda\condabin
C:\Miniconda\envs\foo
[...]

PR #137, opened 2020-12-30
PR: Allow activate-environment to be a path

bollwyvl:
This follows up on #136 where we might want to allow for `activate-environment` to be a `--prefix` rather than a `--name` argument to various conda commands.
 
Also removes some dead code which was using `--name`, and tweaks some more docs and typos in log messages.

Of worrying note: I have now seen at least one `403` from the GitHub API when requesting the miniforge releases, despite my best efforts at caching. We probably need to:
- have a hard-coded "version of last resort" (the one where it dropped msys2 would be a good one :P) 
- throw some weight on https://github.com/conda-forge/miniforge/issues/76
- use `git` to list tags and build URLs from that... but we might pull pre-releases, etc.


goanpeca:
@bollwyvl I did not understand the issues before, querying Gtihub API is expensive as you have now noticed.

Why do we need that?

bollwyvl:
> Why do we need the miniforge releases?

At present, `miniforge-version` is not _required_, though of course we could: demanding a version seems like it might be less nice for the simplest use case. Unless we do, we need to get a version from _somewhere_ as there is no `latest` at a predictable URL. Having the page on the above linked issue would give us that, as would hard-coding a version-of-last-resort... it seems to be getting released more often than miniconda.

goanpeca:
>At present, miniforge-version is not required, though of course we could: demanding a version seems like it might be less nice for the simplest use case.

Might be less nice, but is what you have to provide to overcome the unnecessary problems with the API and havig a page with all that (not happenning soon) so I prefer going with explicit over implicit here :)

jaimergp:
This might not be as nice and convenient for the user but what if we deprecate `miniconda-version`, `miniforge-version`, `miniforge-variant` and friends and just expose `installer`? This could be either a URL to a constructor-derived installer, or a keyword like `Miniconda-4.12` or `Mambaforge-4.12`.

If finding the right installer URL is tricky, we could collect a list in our up and coming fancy docs.

bollwyvl:
> rather have the miniforge-version requirement.

Sure, sounds good. seems out of scope for this PR, will do another... one thing we _can_ do then is have a default `miniforge-variant` of `Miniforge`, so it would still only be a single input to get "vanilla" conda/cpython at a particular version.

goanpeca:
>Sure, sounds good. seems out of scope for this PR, will do another... one thing we can do then is have a default miniforge-variant of Miniforge, so it would still only be a single input to get "vanilla" conda/cpython at a particular version.

Agreed!

--- 

@jaimergp let's explore our options on another issue :-p

goanpeca:
Is this ready for review @bollwyvl ?

bollwyvl:
> deprecate miniconda-version, miniforge-version, miniforge-variant

i feel like deprecation makes people sad... because fixing ci that _used to work_ makes people sad. and the URL structures are fairly different between them, anyhow, even though they both come out of constructor. another reason i like the separate ones is that it becomes easy to use github search for [examples](
https://github.com/search?l=YAML&q=miniconda-version&type=Code) when you have semantically-rich strings.

bollwyvl:
> Is this ready for review

yeah, unless we want to sneak anything else into the tests... i guess I haven't tried giving it a `base` env, might break the world...

bollwyvl:
oh, and @goanpeca, re: #136: can you comment on why we always add `$CONDA/condabin` to path?

bollwyvl:
Looks like we need to keep `condabin`... will revert...

src/conda.ts
```diff
@@ -144,10 +157,7 @@ export async function condaInit(
   options: types.IDynamicOptions
 ): Promise<void> {
   let ownPath: string;
-  const isValidActivate: boolean =
-    inputs.activateEnvironment !== "base" &&
-    inputs.activateEnvironment !== "root" &&
-    inputs.activateEnvironment !== "";
+  const isValidActivate = !utils.isBaseEnv(inputs.activateEnvironment);
```
goanpeca, line 160: is `utils.isBaseEnv` also checking for an empt string env?

goanpeca (approved):
One question, otherwise looks good :) 

Thanks @bollwyvl !

src/conda.ts
bollwyvl:
```

/** Names for a conda `base` env */
export const BASE_ENV_NAMES = ["root", "base", ""];
```

PR #138, opened 2020-12-31
Rework Miniforge inputs

bollwyvl:
- follow-on to #133
- addresses concerns from https://github.com/conda-incubator/setup-miniconda/pull/137#issuecomment-752682346
- [x] stop doing fragile github actions stuff
- [x] make `miniforge-version` the indicator for using `download-miniforge`
- [x] make `Miniforge3` the default `miniforge-variant`
- [x] update docs
- [x] update example 10 test


goanpeca:
Is this ready for review @bollwyvl ?

action.yml
```diff
@@ -20,7 +20,7 @@ inputs:
     description:
       "If provided, this variant of Miniforge will be downloaded and installed.
       Currently-known values:
-      - Miniforge3
+      - Miniforge3 (default)
```
goanpeca, line 23: @bollwyvl we should add the default here no?

```yaml
    required: false
    default: "Miniforge3"	    
```

goanpeca (changes requested):
I think the action yaml needs a small fix.

Otherwise LGTM :)

action.yml
bollwyvl: i'll take a look...

bollwyvl: 510ef46  and 6689c3c

goanpeca (approved):
Awesome! Thanks a lot @bollwyvl 🥳 !

action.yml
bollwyvl: 6689c3c points out that we still have some incidental coupling between the various inputs... it still seems manageable, but i wish i could think of a way to _simply_ express these constraints more declaratively... adding something like JSON schema could do, but more deps...

PR #140, opened 2021-01-02
support miniforge-version: latest, add more architectures

bollwyvl:
Happy New Miniforge :tada: ! Turns out we _can_ get `miniforge-version: latest` at a predictable (though _slightly_ different) URL.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/linking-to-releases

## Questions
- @goanpeca what's the motivation for the specific `arch` inputs on the miniconda architectures? On the miniforge ones, not knowing what to put, i just made it a pass-through, but renamed the constant so we don't get them mixed up...

## Code changes
- [x] supports the `latest` URL scheme
- [x] fills out rest of architectures supported by miniforge (I guess for self-hosted runners) 
- [x] docs
- [x] test

### DIFF
--- a/.github/workflows/example-10.yml
+++ b/.github/workflows/example-10.yml
@@ -28,7 +28,7 @@ jobs:
         os: ["ubuntu", "macos", "windows"]
         include:
           - os: windows
-            miniforge-version: 4.9.2-4
+            miniforge-version: latest
     steps:
       - uses: actions/checkout@v2
       - uses: ./
@@ -67,7 +67,7 @@ jobs:
           - os: macos
             environment-file: etc/example-empty-channels-environment.yml
             miniforge-variant: Mambaforge-pypy3
-            miniforge-version: 4.9.2-4
+            miniforge-version: latest
           - os: windows
             environment-file: etc/example-explicit.Windows.conda.lock
             condarc-file: etc/example-condarc.yml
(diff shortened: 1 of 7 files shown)


goanpeca:
Hi @bollwyvlThe motivation was 32/64 bit installers

bollwyvl:
> the motivation was 32/64 bit installers

Right, but why are they named like they are? As in, `ARM64` vs `aarch64`...

goanpeca:
Cant recall :)

bollwyvl:
Guess it'd good to go! And hooray, no parsing (though the page is coming along over on https://github.com/conda-forge/miniforge/pull/89)

.github/workflows/example-10.yml
```diff
@@ -26,16 +26,13 @@ jobs:
...
-          - os: windows
-            miniforge-version: 4.9.2-4
     steps:
       - uses: actions/checkout@v2
       - uses: ./
         id: setup-miniconda
         with:
           environment-file: etc/example-environment.yml
-          miniforge-version: ${{ matrix.miniforge-version }}
+          miniforge-version: latest
```
jGaboardi, line 35: Should this be `miniforge-version: "latest"` to conform with `miniconda-version: "latest"` or does that not matter here?

bollwyvl: it doesn't really matter EVER, as everything is a string in GHAYAML. my kingdom for some `*` and `&` anchors... we could likely normalize everything to not be quoted.

bollwyvl: the whole `=== 'true'` thing (and a lack of real schema semantics) is really tiresome

goanpeca (approved):
LGTM thanks @bollwyvl :)

PR #146, opened 2021-02-05
Handle unset miniforge-version

### LINKED ISSUE #145: Not setting miniforge-version silently uses bundled miniconda
If you only set `miniforge-variant` but not `miniforge-version`. The bundled `miniconda` is used with the current code on master.

bollwyvl:
Setting `miniforge-variant` but _not_ setting `miniforge-version` doesn't use that variant.

- fixes #145
- [x] test
  - [repro](https://github.com/conda-incubator/setup-miniconda/pull/146/checks?check_run_id=1839861005#step:4:10) on 3da7dcf:  `mamba: command not found`
- [x] fix
- [x] docs

### DIFF
--- a/.github/workflows/example-10.yml
+++ b/.github/workflows/example-10.yml
@@ -65,10 +65,10 @@ jobs:
             environment-file: etc/example-empty-channels-environment.yml
             miniforge-variant: Mambaforge-pypy3
             miniforge-version: latest
+          # should use mamabforge `latest`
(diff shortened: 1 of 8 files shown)


goanpeca (approved):
Thanks for the quick turnaround @bollwyvl :)

LGTM!

bollwyvl:
Thanks ☺️

PR #148, opened 2021-02-06
Remove line preventing the use of 32 bit version on Linux

### LINKED ISSUE #147: using 32-bit versions of miniconda on Linux
I am trying to use this action to set up a 32-bit conda environment for a test. My current options are:
```
  anaconda32:
    name: anaconda-32-bit
    runs-on: ubuntu-latest
    defaults:
      run:
        shell: bash -l {0}  # setup-miniconda requires using login bash shells to activate env
    strategy:
      fail-fast: false
    steps:
      - uses: actions/checkout@v2
      - uses: conda-incubator/setup-miniconda@v2
        with:
          auto-update-conda: true
          miniconda-version: latest
          architecture: x86
      - name: Conda Info
        run: conda info
```
I get [this](https://github.com/ktbarrett/find_libpython/runs/1843620693?check_suite_focus=true) message: `32-bit Linux is not supported by recent versions of Miniconda`.

Reading the [documentation](https://github.com/conda-incubator/setup-miniconda/blob/2fc91a2ef82dfb50ded2d80e56f83b7039ba5955/action.yml#L12-L18) on the `miniconda-version` option, I see that it mentions looking at [this](https://repo.anaconda.com/miniconda/) page for supported Miniconda versions. I guess I'm missing something because `Miniconda3-latest-Linux-x86.sh` exists, so I would think specifying `latest` as the `miniconda-version` should work.

ktbarrett:
Line prevented all use of `architecture: x86` and Linux distributions outright, which is not reasonable. While it is true that 32-bit version of Miniconda have not been produce in a while: 1) This isn't the right way to implement that check, it should be a warning at worst. 2) It shouldn't apply to specific version, only when 'latest' is specified; but it currently applies to *all* versions. ~~3) 'latest' is still the only way to refer to the latest release of Miniconda. There is no version alias for that release (AFAICT, based on dates).~~

Closes #147.

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -61,9 +61,6 @@ const RULES: IRule[] = [
   (i) =>
     !!(i.minicondaVersion && i.architecture !== "x64") &&
     `'architecture: ${i.architecture}' requires "miniconda-version"`,
-  (i) =>
-    !!(i.architecture === "x86" && constants.IS_LINUX) &&
-    `'architecture: ${i.architecture}' is not supported by recent versions of Miniconda`,
 ];
 
 /*


bollwyvl:
Lgtm (not at machine)

Issue #158, opened 2021-03-29
Python 3.7 does not appear to be supported anymore


sam-hoffman:
Hi! My env file looks like this: 

```
name: my-env
channels:
  - conda-forge
  - defaults
dependencies:
  - python<3.8
  - networkit==6.1.0
...
```

However, the `setup-miniconda` logs report that Python 3.8.5 is being added to the env file, which then predictably fails
```
2021-03-29T16:03:37.0343949Z Making patched copy of 'environment-file: my-service/env.yml'
2021-03-29T16:03:37.0344996Z name: my-env
2021-03-29T16:03:37.0345729Z channels:
2021-03-29T16:03:37.0346613Z   - defaults
2021-03-29T16:03:37.0347511Z   - conda-forge
2021-03-29T16:03:37.0348311Z dependencies:
2021-03-29T16:03:37.0349231Z   - python<3.8
2021-03-29T16:03:37.0350128Z   - networkit==6.1.0
...
2021-03-29T16:03:37.0370060Z   - python=3.8.5
2021-03-29T16:03:37.0370462Z 
2021-03-29T16:03:37.0371612Z ##[group]Updating 'my-env' env from conda env update...
2021-03-29T16:03:37.0373261Z [command]/usr/share/miniconda/condabin/conda env update --name my-env --file /tmp/environment-patched.yml
2021-03-29T16:04:09.3426732Z Collecting package metadata (repodata.json): ...working... done
2021-03-29T16:04:09.3822645Z Solving environment: ...working... failed
2021-03-29T16:04:09.3833700Z ##[warning]
ResolvePackageNotFound: 
  - python[version='3.8.5.*,<3.8']
```

Is this related to the new update? It has not failed in the past. I tried pinning `setup-miniconda@2.0.0` but had the same error result. Ideally, I'd like to be able to specify the Python version in my `env.yml` file.

jaimergp:
Can you try leaving your env file unrestrained when it comes to `python` and specify `python-version: 3.7`?

```yaml
name: my-env
channels:
  - conda-forge
  - defaults
dependencies:
  - python
  - networkit==6.1.0
  ....
```

sam-hoffman:
Unfortunately for some of our codebases we need to specify the Python version in the `env.yml` file. 

Would you accept a PR to allow the Python version specified in the `env.yml` file to override the Python version defaulted to in the github action?

jaimergp:
[...]

PR #161, opened 2021-03-29
Copy channels before reversing

### LINKED ISSUE #160: Channel priority changed in 2.1.0 release
Distributed's CI specifies the conda channels to use inside a [conda environment file](https://github.com/dask/distributed/blob/25a2867e8e950d9cb51b071337ffc43ddaf0bb52/continuous_integration/environment-3.8.yaml) where we prefer to use conda-forge where possible and fallback to using defaults when needed.

After the `setup-miniconda` 2.1.0 release we started observing many more packages being installed from defaults instead of conda-forge (see the links below which compare the packages installed using `setup-miniconda=2.0.1` and `setup-miniconda=2.1.0`). Was this an intentional change that requires users to make changes to our GitHub actions workflow? 

- [`conda list` with `setup-miniconda=2.0.1`](https://github.com/dask/distributed/pull/4646/checks?check_run_id=2220556762#step:7:7)
- [`conda list` with `setup-miniconda=2.1.0`](https://github.com/dask/distributed/runs/2220484541#step:7:7)

bollwyvl:
- fixes #160
- [x] use `.slice` to copy channels before reversing
- [x] updates `package-lock.json` after installing from fresh environment
- [x] add "typical" channel line-up to example environment 
- [ ] add test of `environment-patched.yml`? Hoist to output?


bollwyvl:
By _inspection_, this is fixing the issue:
https://github.com/conda-incubator/setup-miniconda/pull/161/checks?check_run_id=2222447371#step:3:257

the order is as set in the original yaml. I don't know how to _simply_ retrieve the logs from within the action (if that's even possible), but otherwise this is ready for review.

bollwyvl:
I guess we could hoist the "effective used environment" to an action output...

jaimergp:
So what was the issue? `.reverse()` acting in place instead of returning a reversed "view" and that can leave the list of channel in the wrong state for the next operation?

bollwyvl:
> acting in place instead of returning a reversed "view"

right, gory details on #160

jaimergp:
Huh, that's surprising indeed, at least coming from the Python world. If that's the case and this fixes the misbehavior, this LGTM.

How would you like for the right behavior? Diffing `conda list`s?

bollwyvl:
Yeah dunno, re: testing.... Maybe something in conda-meta/history?

goanpeca:
Looks good @bollwyvl 

Should I make a 2.1.1 release?

bollwyvl:
Good by me to release unless we can figure out what the other fresh issue
is regarding python-version... #158... Might be a change, but not
necessarily a _bug_ whereas this definitely was.

PR #163, opened 2021-03-31
Create patched YML in place

### LINKED ISSUE #162: Environment file with pip dependency file
It seems like the latest `v2.1.0` has broken the processing of custom environment file with 

```
  - pip:
    - -r file:requirements.txt
```

That now seems to look for it at a non-existing location:

```
  Installing pip dependencies: ...working... Ran pip subprocess with arguments:
  ['/usr/share/miniconda/envs/devito/bin/python', '-m', 'pip', 'install', '-U', '-r', '/tmp/condaenv.zimdqc7e.requirements.txt']
  Pip subprocess output:
  
  failed
  Warning: Pip subprocess error:
  ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
```

Is this supposed to not be supported anymore and what's the proposed fix?

bollwyvl:
## References

- fixes #162

## Changes
- [x] creates patched `setup-miniconda-patched-{environment.yml}` in-place (to restore compatibility with `pip: ["-r requirements.txt"]` (#162) or other cases, e.g. `-e .` (would also break today)
- [x] adds some `environment-file(-content|-was-patched)` outputs about the effective `environment-file` used
- [x] adds `clean-patched-environment-file` (default `true`) because _I_ don't like actions leaving junk i don't ask for
  - [x] adds output message 
- [x] add some tests (can now add real test for #161)
- [x] docs for inputs and outputs
- [x] changelog


action.yml
```diff
@@ -1,6 +1,25 @@
 name: "Setup Miniconda"
 author: Gonzalo Peña-Castellanos (@goanpeca)
 description: "Set up Conda package and environment manager with Miniconda."
+
+runs:
```
bollwyvl, line 5: just moved these up top so they were more discoverable

src/outputs.ts
```diff
@@ -34,3 +34,17 @@ export async function setCacheVariable(options: types.IDynamicOptions) {
...
+ */
+export function setEnvironmentFileOutputs(
+  envFile: string,
+  envContent: string,
+  patched = false
+): void {
+  core.setOutput(constants.OUTPUT_ENV_FILE_PATH, path.resolve(envFile));
+  core.setOutput(constants.OUTPUT_ENV_FILE_CONTENT, envContent);
+  core.setOutput(constants.OUTPUT_ENV_FILE_PATCHED, patched ? "true" : "false");
+  core.saveState(constants.OUTPUT_ENV_FILE_PATCHED, patched);
```
bollwyvl, line 49: gah, i didn't know about `saveState`... should be using that instead of `dynamicOptions` or whatever... though it's all `any`s all the way down.

src/setup.ts
```diff
@@ -87,6 +87,15 @@ async function setupMiniconda(inputs: types.IActionInputs): Promise<void> {
     );
   }
 
+  if (
+    inputs.cleanPatchedEnvironmentFile === "true" &&
+    core.getState(constants.OUTPUT_ENV_FILE_PATCHED)
+  ) {
+    await core.group("Cleaning up patched environment-file...", async () =>
```
bollwyvl, line 94: some future refactor could make this into a "post" step...

bollwyvl:
Tests are up... I don't know how I lose my bash knowledge on a seeming day-to-day basis...

bollwyvl:
gah, didn't add a test for `-r requirements.txt`...

bollwyvl:
Ok, does the `pip: ["-r requirements.txt"]` thing in test now.

src/setup.ts
goanpeca: Mind creating an issue to keep this on the top of our heads @bollwyvl ?

goanpeca (approved):
Thanks for working on this @bollwyvl !

Works like a charm!

goanpeca:
Will cut a new release!

src/setup.ts
bollwyvl: raised https://github.com/conda-incubator/setup-miniconda/issues/165!

goanpeca:
Released 2.1.1 @bollwyvl !

bollwyvl:
Thanks all!

PR #168, opened 2021-04-02
Fix logic in 32-bit architecture check

ktbarrett:
Closes #166.

PR #189, opened 2021-08-09
Fixes #166: Error on miniconda-version *not* specified instead of when it *is* specified.

### LINKED ISSUE #166: Still having errors with 32-bit miniconda setups
From [here](https://github.com/ktbarrett/find_libpython/pull/13/checks?check_run_id=2221834708#step:3:10). I am specifying `miniconda-version` and still get errors. The following line is to blame (I think, I don't know JS), it seems that it should check if miniconda-version is **not** set and architecture != x64, based on the error message.
https://github.com/conda-incubator/setup-miniconda/blob/35d1405e78aa3f784fe3ce9a2eb378d5eeb62169/src/input.ts#L61-L63

chrisjbillington:
A simple fix to what appeared to be a simple bug.

Raise an error if on non-x64 architectures, `miniconda-version` is *not* specified, instead of when it *is* specified.


jaimergp:
Thanks! Do we need a test for this or is it enough with what we have so far?

chrisjbillington:
I thought I'd add that this does not fix the more general problem of old conda builds (many of which are 32-bit) being unsupported - ongoing attempts to address that issue are in #168.

However, some 32-bit builds, namely Windows, [have recent relases](https://repo.anaconda.com/miniconda/):

Miniconda3-latest-Windows-x86.exe | 55.3M | 2021-07-21 11:05:08 | a082be7ef6dd4f14b974dab2d78c5693
-- | -- | -- | --

So this fix restores the ability to use them, at least.

ktbarrett:
I think tests would show that 32 bit support is still not quite there, but I don't think that fact should block this obvious bugfix from coming in.

jaimergp (approved):
LGTM but I'd have @bollwyvl have a look too!

bollwyvl:
Yep, the fix looks good. As it was a bug, we probably do need a test, which can then also be used for documentation purposes in the README.

Also it probably needs to have the build re-run to update the transpiled assets

jaimergp:
This seems to be successfully using x86-latest on Windows now and failing (on purpose) on Ubuntu/Mac. I did this because latest x86-Ubuntu predates 4.6, which is the oldest we are supporting currently.

chrisjbillington:
Thanks very much for fleshing this out @jaimergp! That one-character change was getting close to the limits of what I might be able to do with javascript. Really appreciate it!

ktbarrett:
Any status on this?

jaimergp:
Let's ping core :) @conda-incubator/setup-miniconda

ktbarrett:
3 months later this has been approved, but still hasn't been merged. What gives?

goanpeca:
Whoops, this fell under the cracks! Sorry

Merging˜!

Issue #192, opened 2021-08-17
Disable error reporting if env cannot be solved


jaimergp:
When `conda` cannot solve an environment, it will try to build a conflict report detailing the possible causes. This takes a while and is often not very helpful. Users are mostly recommended to try solve the same environment with `mamba` to get a better report.

Turns out [there's a config key](https://docs.conda.io/projects/conda/en/latest/configuration.html) (check bottom of page) to disable these reports (or _hints_ as they call it): `unsatisfiable_hints`. If set to `false`, `conda` will [not compute them](https://github.com/conda/conda/blob/master/conda/resolve.py#L343). 

Do you think this would be a sane default behaviour for this action? It can save a lot of CI resources down the line!

bollwyvl:
If that also disables the `mamba` report, then that's probably not such a good default, as the report is often quite useful from that backend... but i'll agree that the conda output is all but useless...

jaimergp:
I guess we can check if `use-mamba` is set before applying the new default?

goanpeca:
Hello, is this something we would like to expose? Not changing the default, but just exposing the option?

@bollwyvl @jaimergp ?

goanpeca:
Covered by FORCED_ERRORS behavior

PR #193, opened 2021-08-17
set unsatisfiable_hints=false for conda only

jaimergp:
Comes from #192


jaimergp:
Huh, looks like there's some kind of non-determinism there... The "cleaner implementation" is faulty, so ignore that one.

The two commits there fail in different tests that, otherwise, would pass with an unset `unsatisfiable_hints`. I will look into this.

jaimergp:
Yes, this is known and needs to be fixed at the `conda` level: https://github.com/conda/conda/issues/9862

goanpeca:
@jaimergp could you update with the latest `develop` branch ?

jaimergp:
Oops, this should have been closed some time ago. Enabling that flag makes conda change behaviour, resulting in unsatisfiable solves (but it should be solved!).

PR #200, opened 2021-11-20
Add caching environment instructions

epassaro:
Provided instructions on how to cache deployed environments for Miniforge variants.


epassaro:
Hi @goanpeca, these are my changes.

EwoutH:
Great addition, thanks!

goanpeca:
Hi @epassaro thanks for this.

Could you on top of this add an actual example that  runs like the others?

See https://github.com/conda-incubator/setup-miniconda/tree/master/.github/workflows

epassaro:
Yes, sure.

epassaro:
Added the example, needs approval to run @goanpeca

README.md
```diff
@@ -570,6 +572,73 @@ you may want to
...
+skip this step if you use an environment file product of `conda env export`
+or `conda list --explicit`.
+
+```yaml
+      - name: Set cache date
+        run: echo "DATE=$(date +'%Y%m%d')" >> $GITHUB_ENV
+        
+      - uses: actions/cache@v2
+        with:
+          path: ${{ matrix.prefix }}
```
jaimergp, line 625: This value is usually available through the `CONDA_PREFIX` environment variable set during activation. You can export that to `GITHUB_ENV` and avoid the matrix key, I believe.

README.md
```diff
@@ -570,6 +572,73 @@ you may want to
...
+          key: ${{ matrix.label }}-conda-${{ hashFiles('etc/example-environment-caching.yml') }}-${{ env.DATE }}-${{ env.CACHE_NUMBER }}
```
jaimergp, line 626: To expand on what you have said here (which is on point), we might need to explain how hashing the env file is not the same as hashing the _resolved_ env file. Since you are only freezing for 24h this is not a big deal, but maybe we can expand on what those inconsistencies might be.

README.md
jaimergp: Ah, but it won't be available before the action is run, I see. Hm, let's keep it like it is then.

epassaro: Ok.

README.md
epassaro: I added a more detailed explaination, I don't know if that's enough @jaimergp

README.md
```diff
@@ -570,6 +573,85 @@ you may want to
...
+available on the channels at installation time. Since packages are updated
+all the time, you will not see these changes reflected in the cache until
+the key gets updated by date.
+
+**This means that the same environment file can make your tests pass locally
+but fail on CI, or the other way around. In that case, reset the cache
+manually to see if that leads to consistent results, or use a resolved
+environment file.**
+
+Finally,  update the environment based on the environment file if the cache
```
jaimergp, line 646: Ugh I hate myself sometimes :D  

```suggestion
Finally, update the environment based on the environment file if the cache
```

README.md
jaimergp: It's perfect, thank you so much!

jaimergp:
Thank you so much @epassaro! We can merge once the CI is passing!

jonashaag:
This doesn't seem to work as advertised. The `DATE` env var isn't taken into account:

```
win-64-conda-95d949d6b25a373b614d944993ac0d3be055e69ebfaf5fae91173f00c5b4106d--0
```

There should be a date between the last two `-`.

epassaro:
> This doesn't seem to work as advertised. The `DATE` env var isn't taken into account:
> 
> ```
> win-64-conda-95d949d6b25a373b614d944993ac0d3be055e69ebfaf5fae91173f00c5b4106d--0
> ```
> 
> There should be a date between the last two `-`.

I realized that the unix command `date` doesn't work on Windows. I need to add some logic in order to make it work for all the OSes. Please open an issue.

jonashaag:
See #212

Issue #208, opened 2021-12-17
Timing issue when renaming ~/.bashrc to ~/.profile


martin-g:
I use the action on a self-hosted runner.
The VM is powerful and I run several Github Action runner instances on it.

From time to time the job fails with:
```
Initializing conda shell integration...
  /home/ubuntu/miniconda/condabin/conda init --all
  no change     /home/ubuntu/miniconda/condabin/conda
  no change     /home/ubuntu/miniconda/bin/conda
  no change     /home/ubuntu/miniconda/bin/conda-env
  no change     /home/ubuntu/miniconda/bin/activate
  no change     /home/ubuntu/miniconda/bin/deactivate
  no change     /home/ubuntu/miniconda/etc/profile.d/conda.sh
  no change     /home/ubuntu/miniconda/etc/fish/conf.d/conda.fish
  no change     /home/ubuntu/miniconda/shell/condabin/Conda.psm1
  no change     /home/ubuntu/miniconda/shell/condabin/conda-hook.ps1
  no change     /home/ubuntu/miniconda/lib/python3.8/site-packages/xontrib/conda.xsh
  no change     /home/ubuntu/miniconda/etc/profile.d/conda.csh
  no change     /home/ubuntu/.bashrc
  no change     /home/ubuntu/.zshrc
  no change     /home/ubuntu/.config/fish/config.fish
  no change     /home/ubuntu/.xonshrc
  no change     /home/ubuntu/.tcshrc
  No action taken.
  Renaming "/home/ubuntu/.bashrc" to "/home/ubuntu/.profile"
  
Error: ENOENT: no such file or directory, rename '/home/ubuntu/.bashrc' -> '/home/ubuntu/.profile'
```

I use `remove-profiles: false` but it seems this does not prevent the renaming.

Could this be improved to be more friendly to such self-hosted setups ?

martin-g:
Is there any workaround for this issue ?
Almost half of the job runs fail because of this: 
https://github.com/martin-g/zeppelin/runs/4618065075?check_suite_focus=true

I tried `touch`ing these files before running setup-miniconda but it does not help too: 

https://github.com/martin-g/zeppelin/blob/1a6ffb862a20f28fb66d43ace21b1f9052caa9f6/.github/workflows/core.yml#L63-L74

schudoku:
[...]

PR #209, opened 2021-12-29
Issue #208 - Do not move non-existing files

martin-g:
This is a small improvement to not try to move .bashrc into .profile if
.bashrc does not even exists.

This is needed for non-ephemeral self-hosted Github Actions runners
where one VM could run multiple runners.

There is still a chance another GHA runner to move the file between the `existsSync` and `io.mv` calls.


martin-g:
> There is still a chance another GHA runner to move the file between the `existsSync` and `io.mv` calls.

I don't see a solution for this problem except stop using the shell rc files completely.

Would it be possible to just issue `conda activate XYZ` when/where needed instead of adding `bashExtraText` to `~/.profile` ?

goanpeca:
>Would it be possible to just issue conda activate XYZ when/where needed instead of adding bashExtraText to ~/.profile ?

Unbfortunately it will not work :-\

goanpeca (approved):
LGTM Thanks!

goanpeca:
Sorry for the delays! And thanks again for your contribution 🚀

PR #212, opened 2022-01-19
Fix caching example

jonashaag:
- No need to hardcode paths
- Write date to step output


jonashaag:
@jaimergp mind approving the workflows?

jonashaag:
@jaimergp

jonashaag:
@jaimergp ping

PR #249, opened 2022-10-17
Avoid deprecated set-output command

klieret:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


goanpeca:
Thanks for the contribution @klieret 🚀

PR #251, opened 2022-11-02
Fix typo in README

yuvipanda:


yuvipanda:
Note that the action *is* actually failing in master, it was just not visible before because of the typo https://github.com/conda-incubator/setup-miniconda/actions/runs/3374010040

goanpeca:
Thanks for the fix @yuvipanda :)

PR #265, opened 2022-11-25
ENH: capability to trust bundled conda

rraymondgh:
I'm using https://github.com/nektos/act to run actions locally (actually pre-release capabilities).  The host / runner is a vagrant / ubuntu box where I have installed latest version of mamba forge.  Hence when using **setup-miniconda** action I want to use bundled conda / mamba and *trust* it so that it uses **mamba** if `use-mamba` parameter is set.

To keep with approach used for bundled where `CONDA` environment variable needs to be set, have added a second environment variable `TRUST_BUNDLED` which then means **mamba** will be used if installed and `use-mamba` is set.

### DIFF
--- a/dist/delete/index.js
+++ b/dist/delete/index.js
@@ -3984,13 +3984,14 @@ var __importStar = (this && this.__importStar) || function (mod) {
     return result;
 };
 Object.defineProperty(exports, "__esModule", ({ value: true }));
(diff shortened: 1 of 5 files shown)


rraymondgh:
Not sure why there is a linting failure, have run everything in contributing guide.  Have not updated README (maybe should)

goanpeca:
Hi @rraymondgh thanks for the PR. I have been a bit busy these past days and will get to this next week

Cheers and sorry for the delays!

goanpeca:
@conda-incubator/setup-miniconda any thoughts on this addition?

PR #289, opened 2023-07-26
Consider leading with conda activation does not work on sh, please use bash

ras44:


goanpeca (approved):
Thanks for your PR @ras44 🚀

PR #293, opened 2023-10-16
Remove Python 2.7 from test matrix (EOL since  April 2020, >4 years)

dbast:
Add Python 3.11 instead as upper bound test (can be upgraded to 3.12 soon).


goanpeca (approved):
Hi @dbast, thanks for the PR, looks good 🚀

goanpeca (approved):

PR #296, opened 2023-10-19
Update Miniconda architectures

marcoesters:
The supported Miniconda architectures are outdated and do not cover all available installers. In particular, with the addition of M1 runners, `osx-arm64` installers need to be supported.

CI failures appear to be unrelated to this PR.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -235,8 +235,8 @@ inputs:
     default: ""
   architecture:
     description:
-      'Architecture of Miniconda that should be installed. Default is "x64".
-      The CPU architecture of the runner is not detected by the workflow.'
+      'Architecture of Miniconda that should be installed. Default is "x64". The
+      CPU architecture of the runner is not detected by the workflow.'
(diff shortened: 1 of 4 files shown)


action.yml
```diff
@@ -235,8 +235,9 @@ inputs:
     default: ""
   architecture:
     description:
-      'Architecture of Miniconda that should be installed. Available options on
-      GitHub-hosted runners are "x86" and "x64". Default is "x64".'
+      'Architecture of Miniconda that should be installed. Default is "x64". It
+      is recommended to provide this value because the CPU architecture of the
+      runner is not detected by the workflow.'
```
mattwthompson, line 240: I don't think recommending users specify this is necessary - the default is fine for a vast majority of workflows

mattwthompson: I think that's a better change

goanpeca (approved):
Thanks for the fix !

PR #299, opened 2023-11-06
Fix condaBasePath when useBundled is false, and there's no pre-existing conda

isuruf:
Useful when using setup-miniconda with self-hosted runners without conda installed.

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -47248,14 +47248,12 @@ const utils = __importStar(__nccwpck_require__(1314));
  * Provide current location of miniconda or location where it will be installed
  */
 function condaBasePath(options) {
-    let condaPath = constants.MINICONDA_DIR_PATH;
-    if (!options.useBundled) {
-        if (constants.IS_MAC) {
-            condaPath = "/Users/runner/miniconda3";
-        }
-        else {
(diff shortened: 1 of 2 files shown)


goanpeca (changes requested):
@isuruf why did you remove the code for the mac case? 

It seems like this change will break something. Could you confirm 🤔 ?

isuruf:
It's the same as `path.join(os.homedir(), "miniconda3");` right?

goanpeca:
>It's the same as path.join(os.homedir(), "miniconda3"); right?

I think not, and home dir might have been pointing somewhere else on mac, hence the hardcoded path. Could you please verify?

isuruf:
With `macos-latest` image it is the home directory.

isuruf:
Ping on this

dbast:
How about running:
```
conda create -n github-action nodejs=16 -c conda-forge
conda activate github-action
npm ci
npm run build
```

and committing the resulting updates so that the example runs actually test the changes proposed by this PR?

goanpeca (approved):

goanpeca (approved):

isuruf:
Thanks @goanpeca

PR #303, opened 2023-11-21
Bring develop changes to main (except latest failing commit)

dbast:
Would like to bring the current changes on develop to main... and then do a bunch of PRs against main directly ... PRs against develop currently don't make sense as the self hosted runners are tested there.

This PR brings all the commits from develop to main, except the last develop commit where testing the self hosted runners started, which currently fails.


dbast:
closing. first getting develop green again.

PR #307, opened 2023-11-22
Run dependabot against main branch and also update node packages

dbast:
To avoid failures when there is no develop branch and issues where dependabot analyses the main branch, but then tries to open a PR against develop, which maybe is already updated.

This should also help to resolve security findings in node packages.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,14 @@ version: 2
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
-    target-branch: "develop"
+    target-branch: "main"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


goanpeca:
Hi @dbast, I do n ot follow the 

"To avoid failures when there is no develop branch", we do all development against that branch.

goanpeca:
> @goanpeca Dependabot is currently failing as after the merge to main the develop branch was deleted (by accident?).


Well it should not have been deleted...

dbast:
restoring it is easy.. thats not the problem... see the other arguments.

goanpeca:
Let's chat over at Element please, I think there is a misunderstanding of how the contribution process works

goanpeca (approved):

PR #308, opened 2023-11-22
Update to node20

dbast:
node16 is eol, thus updating. See also https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

Many popular actions (like checkout) already updated and everything should work fine.

The entire package-lock.json was regenerated via node20.

Also explicitly setting the node version in the lint action.


goanpeca:
Looks good 🚀 , thanks!

goanpeca (approved):

PR #315, opened 2023-11-23
Split lint into check and build, switch from `npm install` to `npm ci`

dbast:
This splits the lint flow into two jobs:
* one only calling `npm run check`, which can just reveal some unrelated formatting changes missing. Those findings still have to be fixed, but in most cases have no consequences in regards to if the example workflow runs are all valid or not.
* one calling `npm run build` and checking if there are no updates = all updates in the ./dist folder are committed... If this job fails, then all the examples runs have to be taken with a grain of salt, as they run agains an outdated ./dist folder, thus being invalid!

This information was previously not visible / masked as mostly `npm run check` failed and prevented `npm run build` from running ... it was then unclear, if that has consequences to the ncc compiled stuff in ./dist or not.

Also switching `npm install` to `npm ci` to install the exact versions from package-lock.json (managed by dependabot). This is faster, more secure and 100% predictable. Thus removing the cron triggers as those won't find anything new.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,27 +9,44 @@ on:
       - "develop"
       - "main"
       - "master"
-  schedule:
-    # Note that cronjobs run on master/main by default
-    - cron: "0 0 * * *"
 
 jobs:
   lint:
-    # prevent cronjobs from running on forks
-    if:
-      (github.event_name == 'schedule' && github.repository ==
-      'conda-incubator/setup-miniconda') || (github.event_name != 'schedule')
     name: Lint
     runs-on: "ubuntu-latest"
-    defaults:
-      run:
-        shell: bash -el {0}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
           node-version: 20
       - run: |
-          npm install
+          # ci: install exact versions from package-lock.json
+          #     fast, secure, predictable compared to npm install
+          npm ci
           npm run check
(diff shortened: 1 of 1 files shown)


dbast:
yay... lint and build each only take 14 sec.

.github/workflows/lint.yml
```diff
@@ -9,27 +9,44 @@ on:
...
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
           node-version: 20
       - run: |
-          npm install
+          # ci: install exact versions from package-lock.json
+          #     fast, secure, predictable compared to npm install
+          npm ci
```
goanpeca, line 25: Should we test both then... `npm ci` and `npm install`  ?

dbast: well, there is no point in doing so... the package-lock.json is basically the solved result of `npm install` and contains all dependencies + transitive dependencies and their sha512 checksums.... but npm is not the subject under test here... we want to get all the dependencies + dev dependencies installed in a predictable+secure way (the hashes are verified) to then run linters / build checks... thats exactly what `npm ci` exists for ... dependabot keeps the package.json + package-json.lock updated / in sync for us.

goanpeca: Got it :)

goanpeca (approved):

goanpeca:
Awesome, thanks for working on this 🚀

PR #316, opened 2023-11-24
Faster workflow response / saving resources via timeout/concurrency policy

dbast:
All the workflows normally finish within <5 minutes. If they take longer (there are several runs >30 minutes in the action history) they always fail and the causes are temporary issues with CDN / dependencies not solvable etc. There is no point then continuing those runs as they anyway fail. Thus setting an upper limit of 30 minutes to save some free Action minutes in the incubator org.

Also setting a concurrency policy for all workflows for faster PR responses when doing repeated pushes to PRs, inspired by [conda](https://github.com/conda/conda/blob/main/.github/workflows/tests.yml#L21).

### DIFF
--- a/.github/workflows/caching-envs-example.yml
+++ b/.github/workflows/caching-envs-example.yml
@@ -13,6 +13,11 @@ on:
     # Note that cronjobs run on master/main by default
     - cron: "0 0 * * *"
 
+concurrency:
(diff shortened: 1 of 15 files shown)


goanpeca (approved):
Awesome! Thanks 🚀

PR #328, opened 2023-12-08
Make conda’s package directories configurable

jezdez:
This adds the ability to use `D:` instead of `C:` on GitHub hosted Windows runners, to increase the speed of decompression.

See https://github.com/conda/conda-build/pull/5082#issuecomment-1845137430 for more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,6 +637,45 @@ jobs:
           use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
 ```
 
+You may also set conda's packages directories (`CONDA_PKGS_DIRS`) config value,
+if you want to change setup-miniconda's default of `~/conda_pkgs_dir` with the
+`pkgs_dirs` config option. This is a comma-separated string like the channels
+config option:
+
+```yaml
+jobs:
+  caching-example:
+    name: Caching
+    runs-on: "windows-latest"
+    steps:
+      - uses: actions/checkout@v4
+      - name: Cache conda
+        uses: actions/cache@v3
+        env:
+          # Increase this value to reset cache if etc/example-environment.yml has not changed
+          CACHE_NUMBER: 0
+        with:
+          # Use faster GNU tar
+          enableCrossOsArchive: true
+          path: D:\conda_pkgs_dir
+          key:
+            ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{
+            hashFiles('etc/example-environment.yml') }}
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+          activate-environment: anaconda-client-env
+          channel-priority: strict
+          environment-file: etc/example-environment.yml
+          pkgs_dirs: D:\conda_pkgs_dir
+```
+
+> Note:
+>
+> - GitHub hosted Windows runners are currently faster during cache
+>   decompression when configuring the packages directoies on the `D:` drive as
+>   shown above. Make sure to use the `enableCrossOsArchive` cache config option
+>   as well.
+
 If you are using pip to resolve any dependencies in your conda environment then
 you may want to
 [cache those dependencies separately](https://docs.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#caching-dependencies),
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -637,6 +637,45 @@ jobs:
...
+          path: D:\conda_pkgs_dir
+          key:
+            ${{ runner.os }}-conda-${{ env.CACHE_NUMBER }}-${{
+            hashFiles('etc/example-environment.yml') }}
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+          activate-environment: anaconda-client-env
+          channel-priority: strict
+          environment-file: etc/example-environment.yml
+          pkgs_dirs: D:\conda_pkgs_dir
```
goanpeca, line 669: This should be `pkgs-dirs` :)

README.md
```diff
@@ -637,6 +637,45 @@ jobs:
...
+```
+
+> Note:
+>
+> - GitHub hosted Windows runners are currently faster during cache
+>   decompression when configuring the packages directoies on the `D:` drive as
```
goanpeca, line 675: typo: directoies

goanpeca:
Thanks for this work @jezdez !

Left a couple of minor comments 🚀

README.md
jezdez: Good catch!

README.md
```diff
@@ -637,6 +637,45 @@ jobs:
           use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
 ```
 
+You may also set conda's packages directories (`CONDA_PKGS_DIRS`) config value,
+if you want to change setup-miniconda's default of `~/conda_pkgs_dir` with the
+`pkgs_dirs` config option. This is a comma-separated string like the channels
```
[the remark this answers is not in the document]
jezdez: Yup!

src/conda.ts
```diff
@@ -130,9 +130,22 @@ export async function applyCondaConfiguration(
...
   }
 
+  // Package directories are also comma-separated, like channels
+  let pkgsDirs = utils.parsePkgsDirs(inputs.condaConfig.pkgs_dirs);
+  for (const pkgsDir of pkgsDirs) {
+    core.info(`Adding pkgs_dir '${pkgsDir}'`);
+    await condaCommand(["config", "--add", "pkgs_dirs", pkgsDir], options);
+  }
+  // We're also setting the appropriate conda config environment variable, to be safe
+  core.exportVariable("CONDA_PKGS_DIRS", pkgsDirs.join());
```
kenodegard, line 140:
```suggestion
  core.exportVariable("CONDA_PKGS_DIRS", pkgsDirs.join(","));
```

src/delete.ts
```diff
@@ -5,34 +5,39 @@ import * as path from "path";
...
  * Clean up the conda cache directory
  */
 async function run(): Promise<void> {
   try {
-    const cacheFolder = utils.cacheFolder();
-
-    if (fs.existsSync(cacheFolder) && fs.lstatSync(cacheFolder).isDirectory()) {
+    const inputs = await core.group("Gathering Inputs...", input.parseInputs);
+    let pkgsDirs = utils.parsePkgsDirs(inputs.condaConfig.pkgs_dirs);
+    if (pkgsDirs.length) {
```
kenodegard, line 18: Can we invert this and exit early instead of deeply nesting everything?

```suggestion
    if (!pkgsDirs.length) return;
```

jezdez:
@conda-incubator/setup-miniconda any takers for another review?

jaimergp:
Code looks good!

Are there any issues with cross-drive hardlinking? I think Windows doesn't like creating those junctions over different units (or if the underlying drive is different), so maybe this is ends up becoming a performance hit _if_ the env is not in the same drive.

So, I wonder if:

- Should we also allow setting `envs-dirs` so users can request that new environments are created in the same drive?
- What about the default path for the installation when conda is bootstrapped from Miniconda or Miniforge installers? Should that also be configurable so we can have _everything_ in the same drive?

jezdez:
I don't know why npm is adding all that cruft to dist/delete/index.js :(

dbast:
> I don't know why npm is adding all that cruft to dist/delete/index.js :(

Afaik the reason is: every new function used from a dependency is copied into the index.js (`npm run build` = ncc = Compile a Node.js project into a single file)

jaimergp (approved):

PR #331, opened 2024-01-31
Enable ARM64 on macOS for Miniforge and Mambaforge.

### LINKED ISSUE #312: miniforge-version support for ARM64
Dear all,
I'm trying to setup miniforge and to use mamba on a arm64 runner on Github Workflow.
Here my setup:

image

However the ARM64 arch is not supperted when used with the miniforge-version keyword (while with miniconda it's possible).

By the way I'm asking this because my miniconda setup suddenly stopped to work giving me this error:

image

Thanks for your help!

jezdez:
Closes #312.

See https://github.com/jezdez/setup-miniconda-test/actions for examples.

### DIFF
--- a/.github/workflows/example-13.yml
+++ b/.github/workflows/example-13.yml
@@ -1,0 +1,55 @@
+name: "Example 13: ARM64 for Miniforge and Mambaforge"
+
+on:
+  pull_request:
+    branches:
+      - "*"
+  push:
+    branches:
+      - "develop"
+      - "main"
+      - "master"
+  schedule:
+    # Note that cronjobs run on master/main by default
+    - cron: "0 0 * * *"
+
+jobs:
+  example-13:
+    # prevent cronjobs from running on forks
+    if:
+      (github.event_name == 'schedule' && github.repository ==
+      'conda-incubator/setup-miniconda') || (github.event_name != 'schedule')
+    name: Ex13 (os=${{ matrix.os }} variant=${{ matrix.variant }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["macos-14"]
+        variant: ["Miniforge3", "Mambaforge"]
+        version: ["latest"]
+        architecture: ["arm64", "ARM64"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./
+        id: setup-miniconda
+        continue-on-error: true
+        with:
+          auto-update-conda: true
+          architecture: ${{ matrix.architecture }}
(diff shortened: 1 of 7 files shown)


.github/workflows/example-13.yml
```diff
@@ -0,0 +1,57 @@
...
+      Ex13 (os=${{ matrix.os }} variant=${{ matrix.variant }} version=${{
+      matrix.version }} architecture=${{ matrix.architecture }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["macos-14"]
+        variant: ["Miniforge3", "Mambaforge"]
+        version: ["latest"]
+        architecture: ["arm64", "ARM64"]
```
jaimergp, line 32: What happens if we don't set this up? I'd like to see what happens with a naive call to `setup-miniconda` for Miniforge. Aka this:

```yaml
      - uses: ./
        id: setup-miniconda
        continue-on-error: true
        with:
          miniforge-version: latest
```

jezdez: I added an example

dbast:
The README lists the status of all workflows that run via cron to give an overview that everything works over time ... can we please also add the new example there. thanks!

.github/workflows/example-13.yml
```diff
@@ -0,0 +1,124 @@
...
+      fail-fast: false
+      matrix:
+        os: ["macos-14"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./
+        id: setup-miniconda
+        continue-on-error: true
+        with:
+          miniforge-version: latest
```
dbast, line 104: This installs https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-MacOSX-x86_64.sh 

I would expect as user that the native architecture is correctly detected as osx-arm64 and the according installer is used ...

.github/workflows/example-13.yml
```diff
@@ -0,0 +1,124 @@
...
+        shell: bash -el {0}
+        run: conda create -n unused --dry-run zlib
+      - name: Run mamba
+        shell: bash -el {0}
+        run: mamba --version
+      - name: Check arm64
+        shell: bash -el {0}
+        run:
+          python -c "import platform; assert platform.machine() == 'arm64',
+          platform.machine()"
```
dbast, line 59: Those steps fail when the "if" in line 32 evaluates to false, thus:
```suggestion
```

dbast (approved):

dbast:
@jaimergp Ready for another review.

jaimergp (approved):

PR #301, opened 2023-11-16
Add self-hosted runners for testing

jaimergp:
WIP.

PR #342, opened 2024-04-11
Add `installation-dir` to customize where the installers are installed to

isuruf:
This is useful for self-hosted runners where we can change the installation directory to the workspace directory to allow automatic cleaning up of the miniforge/miniconda environments.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -29,6 +29,11 @@ inputs:
       installers"
     required: false
     default: ""
+  install-dir:
+    description:
+      "If provided, the installer will be installed in the given directory."
+    required: false
+    default: ""
   miniconda-version:
     description:
       "If provided, this version of Miniconda3 will be downloaded and installed.
--- a/.github/workflows/example-10.yml
+++ b/.github/workflows/example-10.yml
@@ -45,7 +45,7 @@ jobs:
           conda list
           python -VV
           printenv | sort
-  
+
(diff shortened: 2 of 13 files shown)


isuruf:
cc @finnagin

action.yml
```diff
@@ -29,6 +29,11 @@ inputs:
       installers"
     required: false
     default: ""
+  install-dir:
```
jaimergp, line 32: I'm inclined to using `installation-directory` here instead. The rest of the options in this project seem to be using no abbreviations (e.g. we fully spell `environment` instead of `env`).

jaimergp (commented):
Just one comment about the name chosen for now. Implementation looks ok. We should add an example test to make sure this works as intended.

action.yml
goanpeca: Hi @isuruf, indeed I agree with @jaimergp comment, we should try to use full names :)

goanpeca (dismissed):
Thanks for working on this @isuruf 

Could we add information on this usage on the README?

Also could we add some sort of workflow to test this?

If this only makes sense for self hosted runners, maybe we could add something on this PR https://github.com/conda-incubator/setup-miniconda/pull/301 ?

jaimergp (approved):

jaimergp:
Thanks!

PR #345, opened 2024-04-23
Investigate errors on missing `$CONDA` on `macos-13`

jaimergp:
Seen in https://github.com/conda/conda-standalone/actions/runs/8799309039/job/24148102977?pr=67 and https://github.com/conda/conda-build/actions/runs/8799101015/job/24147451008


jaimergp:
Yoohoo, reproduces now! We were not considering that GH-hosted x64 runners on macOS could happen. But that's the problem with macos-13. It's x64 and doesn't bundle any $CONDA.

I'll submit a potential fix now.

src/installer/bundled-miniconda.ts
```diff
@@ -15,7 +16,8 @@ export const bundledMinicondaUser: types.IInstallerProvider = {
       inputs.miniforgeVariant === "" &&
       inputs.miniforgeVersion === "" &&
       inputs.architecture === "x64" &&
-      inputs.installerUrl === ""
+      inputs.installerUrl === "" &&
+      MINICONDA_DIR_PATH.length > 0
```
goanpeca, line 20: That list of `&&s` is getting a bit unwieldy 😆 !

goanpeca (approved):
Awesome ! fix looks nice and simple :) 

Merging!

PR #360, opened 2024-07-24
Start deprecation of `miniforge-variant: Mambaforge`

jaimergp:
Comes from https://github.com/conda-forge/miniforge/issues/602

Mambaforge and Miniforge are equivalent these days. Maintaining Mambaforge adds unnecessary burden on the Miniforge repo and creates confusion for end users. We'd like to sunset it eventually, and to do so we are starting the deprecation cycle with a warning.

### DIFF
--- a/.github/workflows/caching-envs-example.yml
+++ b/.github/workflows/caching-envs-example.yml
@@ -36,10 +36,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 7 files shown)


dbast (approved):

jaimergp:
Thanks @dbast!

I'd wait til https://github.com/conda-forge/conda-forge.github.io/pull/2242 is published, then we can merge.

Issue #361, opened 2024-07-25
Nasty edge case in the cached env docs


liamhuber:
I ran into a case where specifying a range of `python` in the env used for the `mamba env update -n anaconda-client-env -f etc/example-environment-caching.yml` step of [the caching docs](https://github.com/conda-incubator/setup-miniconda?tab=readme-ov-file#caching-environments) results in overriding the requested version of python passed to the `setup-miniconda` `python-version` action arg. In a hindsight-is-20/20 moment, it makes perfect sense that the update overrides things an bumps python to the maximum version specified in the range!

It's possible there are a few other manifestations of this symptom; at some point we managed to get `python` and `python_abi` having different versions due to this issue.

The solution I found is to instead take a five-step process so that the cache-free invocation of `setup-miniconda` always exploits the `environment-file` argument, which allows the `python-version` to be respected. Basically:
- Check for a cache with `lookup-only = true`
- If cache hit: run `setup-miniconda` without `environment-file`
- If cache hit: restore cache
- if cache not hit: run `setup-miniconda` _with_ `environment-file`
- if cache not hit: save cache

I made a [repo demonstrating this](https://github.com/liamhuber/setup-miniconda_caching), and as long as GitHub holds the run report, you can [look and see](https://github.com/liamhuber/setup-miniconda_caching/actions/runs/10100287511) that the docs approach has python upgrade while the careful approach outlined above retains the requested version.

Issue #186, opened 2021-07-30
defauls packages are always installed regardless of the channels option


traversaro:
I tipically use `setup-miniconda` to install packages from two or three channels, `conda-forge`, `robotology` or `robostack`. I never need/want to install packages from defaults. I tought that by using the `channels` options as shown in https://github.com/conda-incubator/setup-miniconda#example-4-conda-options the `defaults` channels should have low priority by default, but even in that case, if i try to install the `compilers` meta-package, packages from defaults are installed:

See https://github.com/traversaro/setup-miniconda-always-uses-default-mre/runs/3199430212 (related job: https://github.com/traversaro/setup-miniconda-always-uses-default-mre/blob/main/.github/workflows/install-compilers-only-conda-forge-based-on-ex4.yml) for an example of this:
~~~
2021-07-30T06:55:37.1491185Z   + _libgcc_mutex                      0.1  conda_forge    conda-forge/linux-64       3 KB
2021-07-30T06:55:37.1492319Z   + _openmp_mutex                      4.5  1_gnu          conda-forge/linux-64      22 KB
2021-07-30T06:55:37.1499003Z   + binutils                        2.36.1  hdd6e379_2     conda-forge/linux-64      27 KB
2021-07-30T06:55:37.1500414Z   + binutils_impl_linux-64          2.36.1  h193b22a_2     conda-forge/linux-64      10 MB
2021-07-30T06:55:37.1501953Z   + binutils_linux-64                 2.36  hf3e587d_33    conda-forge/linux-64      22 KB
2021-07-30T06:55:37.1503097Z   + bzip2                            1.0.8  h7f98852_4     conda-forge/linux-64     Cached
2021-07-30T06:55:37.1504025Z   + c-ares                          1.17.1  h7f98852_1     conda-forge/linux-64     Cached
2021-07-30T06:55:37.1505036Z   + c-compiler                       1.2.0  h7f98852_0     conda-forge/linux-64       6 KB
2021-07-30T06:55:37.1506366Z   + ca-certificates              2021.5.30  ha878542_0     conda-forge/linux-64     Cached
2021-07-30T06:55:37.1507325Z   + cmake                           3.21.1  h8897547_0     conda-forge/linux-64      15 MB
2021-07-30T06:55:37.1508214Z   + compilers                        1.2.0  ha770c72_0     conda-forge/linux-64       5 KB
[...]

PR #364, opened 2024-08-26
Remove defaults channel on input `channel: nodefaults`

### LINKED ISSUE #207: `defaults` channel in `.condarc` file when only `conda-forge` is specified
We're using `setup-miniconda` over in [`dask-bigquery`](https://github.com/coiled/dask-bigquery) and discovered that when using [this environment file](https://github.com/coiled/dask-bigquery/blob/c3726943e4b8154744d8fa1c1ef36008a36cefbf/ci/environment-3.9.yaml) which specifies only the `conda-forge` channel, along with [this `conda-incubator/setup-miniconda` GitHub actions setup](https://github.com/coiled/dask-bigquery/blob/c3726943e4b8154744d8fa1c1ef36008a36cefbf/.github/workflows/tests.yml#L32-L42), which is using mambaforge and strict channel priority, we still end up with a few packages installed from the `main` channel. This channel mixing is, I think, leading to CI errors for the project. 

If I look at the [logs](https://pipelines.actions.githubusercontent.com/gIH8ISHGoWJmhfujoO1fwvmvaoltIeZmToyjJAUZ0UbXqwyUdk/_apis/pipelines/1/runs/172/signedlogcontent/10?urlExpires=2021-12-17T01%3A50%3A10.9061058Z&urlSigningMethod=HMACV1&urlSignature=llLPC8pEx3Q7O18xPeunh6PXvG4vXo4vh5zZRV3X1%2B8%3D) from [this CI build](https://github.com/coiled/dask-bigquery/runs/4542574053?check_suite_focus=true)

```
2021-12-16T02:46:39.5858090Z [command]/Users/runner/miniconda3/condabin/conda config --show-sources
2021-12-16T02:46:39.7591180Z ==> /Users/runner/miniconda3/.condarc <==
2021-12-16T02:46:39.7592030Z channels:
2021-12-16T02:46:39.7593320Z   - conda-forge
2021-12-16T02:46:39.7593660Z 
2021-12-16T02:46:39.7594130Z ==> /Users/runner/.condarc <==
2021-12-16T02:46:39.7594690Z auto_update_conda: False
2021-12-16T02:46:39.7595240Z auto_activate_base: False
2021-12-16T02:46:39.7595800Z notify_outdated_conda: False
2021-12-16T02:46:39.7596340Z changeps1: False
2021-12-16T02:46:39.7596820Z pkgs_dirs:
2021-12-16T02:46:39.7597760Z   - /Users/runner/conda_pkgs_dir
2021-12-16T02:46:39.7598350Z channel_priority: strict
2021-12-16T02:46:39.7598860Z channels:
2021-12-16T02:46:39.7599710Z   - conda-forge
2021-12-16T02:46:39.7600550Z   - defaults
2021-12-16T02:46:39.7601040Z always_yes: True
```

it appears `conda config --show-sources`, which is run during `setup-miniconda` shows, two separate sources:

- ` /Users/runner/miniconda3/.condarc`: which only specifies the `conda-forge` channel
- `/Users/runner/.condarc`: which specifies both `conda-forge` and `defaults`

So the `defaults` channel is being added at some point, but it's not clear to me where and how I can disable it from being added. Any help is appreciated

liamhuber:
Changes conda setup to run `conda config --remove channels defaults` when `nodefaults` is found in the `channels` input instead of adding the `- nodefaults` channel. I.e.

```
- uses: conda-incubator/setup-miniconda
      with:
        channels: conda-forge,nodefaults
        channel-priority: strict
```

Will correctly give

```
channels:
- conda-forge
dependencies:
...
```

Instead of 

```
channels:
- conda-forge
- nodefaults
dependencies:
...
```

[Here is a log](https://github.com/liamhuber/test-setup-minconda/actions/runs/10568363909/job/29279169033) of the modified version in action.

Closes #207

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -47289,7 +47289,7 @@ function applyCondaConfiguration(inputs, options) {
                 try {
                     yield condaCommand(["config", "--remove", "channels", "defaults"], options);
                 }
-                catch (error) {
+                catch (err) {
                     core.info("Removing defaults raised an error -- it was probably not present.");
                 }
             }
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -47284,9 +47284,14 @@ function applyCondaConfiguration(inputs, options) {
         // LIFO: reverse order to preserve higher priority as listed in the option
         // .slice ensures working against a copy
         for (const channel of channels.slice().reverse()) {
+            core.info("Removing channel defaults");
             if (channel === "nodefaults") {
-                core.info("Removing channel defaults");
-                yield condaCommand(["config", "--remove", "channels", "defaults"], options);
+                try {
+                    yield condaCommand(["config", "--remove", "channels", "defaults"], options);
+                }
+                catch (error) {
+                    core.info("Removing defaults raised an error -- it was probably not present.");
+                }
             }
             else {
                 core.info(`Adding channel '${channel}'`);
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -47284,8 +47284,8 @@ function applyCondaConfiguration(inputs, options) {
         // LIFO: reverse order to preserve higher priority as listed in the option
         // .slice ensures working against a copy
         for (const channel of channels.slice().reverse()) {
-            core.info("Removing channel defaults");
             if (channel === "nodefaults") {
+                core.info("Removing channel defaults");
                 try {
                     yield condaCommand(["config", "--remove", "channels", "defaults"], options);
                 }
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -47284,8 +47284,14 @@ function applyCondaConfiguration(inputs, options) {
         // LIFO: reverse order to preserve higher priority as listed in the option
         // .slice ensures working against a copy
         for (const channel of channels.slice().reverse()) {
-            core.info(`Adding channel '${channel}'`);
-            yield condaCommand(["config", "--add", "channels", channel], options);
+            if (channel === "nodefaults") {
+                core.info("Removing channel defaults");
+                yield condaCommand(["config", "--remove", "channels", "defaults"], options);
+            }
+            else {
+                core.info(`Adding channel '${channel}'`);
+                yield condaCommand(["config", "--add", "channels", channel], options);
+            }
         }
         // All other options are just passed as their string representations
         for (const [key, value] of configEntries) {
--- a/src/conda.ts
+++ b/src/conda.ts
@@ -133,7 +133,7 @@ export async function applyCondaConfiguration(
           ["config", "--remove", "channels", "defaults"],
           options,
         );
-      } catch (error) {
+      } catch (err) {
         core.info(
           "Removing defaults raised an error -- it was probably not present.",
         );
--- a/src/conda.ts
+++ b/src/conda.ts
@@ -126,12 +126,18 @@ export async function applyCondaConfiguration(
   // LIFO: reverse order to preserve higher priority as listed in the option
   // .slice ensures working against a copy
   for (const channel of channels.slice().reverse()) {
+    core.info("Removing channel defaults");
     if (channel === "nodefaults") {
-      core.info("Removing channel defaults");
-      await condaCommand(
-        ["config", "--remove", "channels", "defaults"],
-        options,
-      );
+      try {
+        await condaCommand(
+          ["config", "--remove", "channels", "defaults"],
+          options,
+        );
+      } catch (error) {
+        core.info(
+          "Removing defaults raised an error -- it was probably not present.",
+        );
+      }
     } else {
       core.info(`Adding channel '${channel}'`);
       await condaCommand(["config", "--add", "channels", channel], options);
--- a/src/conda.ts
+++ b/src/conda.ts
@@ -126,8 +126,8 @@ export async function applyCondaConfiguration(
   // LIFO: reverse order to preserve higher priority as listed in the option
   // .slice ensures working against a copy
   for (const channel of channels.slice().reverse()) {
-    core.info("Removing channel defaults");
     if (channel === "nodefaults") {
+      core.info("Removing channel defaults");
       try {
         await condaCommand(
           ["config", "--remove", "channels", "defaults"],
--- a/src/conda.ts
+++ b/src/conda.ts
@@ -126,8 +126,16 @@ export async function applyCondaConfiguration(
   // LIFO: reverse order to preserve higher priority as listed in the option
   // .slice ensures working against a copy
   for (const channel of channels.slice().reverse()) {
-    core.info(`Adding channel '${channel}'`);
-    await condaCommand(["config", "--add", "channels", channel], options);
+    if (channel === "nodefaults") {
+      core.info("Removing channel defaults");
+      await condaCommand(
+        ["config", "--remove", "channels", "defaults"],
+        options,
+      );
+    } else {
+      core.info(`Adding channel '${channel}'`);
+      await condaCommand(["config", "--add", "channels", channel], options);
+    }
   }
 
   // All other options are just passed as their string representations


liamhuber:
The failure of 6 is by design:

```
 CondaKeyError: 'channels': 'defaults' is not in the 'channels' key of the config file
  
  
  
  CondaKeyError: 'channels': 'defaults' is not in the 'channels' key of the config file
  
Error: The process '/home/runner/miniconda3/condabin/conda' failed with exit code 1
```

liamhuber:
Example 6 during `Adding tools to 'base' env...`:

```
  Removing channel defaults
  /home/runner/miniconda3/condabin/conda config --remove channels defaults
  Warning: 
  CondaKeyError: 'channels': 'defaults' is not in the 'channels' key of the config file
  
  
  
  CondaKeyError: 'channels': 'defaults' is not in the 'channels' key of the config file
```

The patch is working as-intendend and in `Applying initial configuration...` we correctly get:

```
  Removing channel defaults
  /home/runner/miniconda3/condabin/conda config --remove channels defaults
```

But then there's something in the Ex6 config that differs from [my simple test case](https://github.com/liamhuber/test-setup-minconda/blob/main/.github/workflows/nodefaults-patch.yml) that causes it to double-dip for `defaults`.

liamhuber:
Ok, got it: the setup [runs `applyCondaConfiguration`](https://github.com/conda-incubator/setup-miniconda/blob/e5293c8fd23bf30f6ad37a6281c8a924c8938440/src/base-tools/index.ts#L62) a _second_ time. So it's just my own new if-clause come to bite me.

liamhuber:
Ok, the log shows the `nodefaults` option correctly purging the `defaults` channel. After that, `defaults` gets added back in because it's included explicitly in `etc/example-environment.yml`. This is a bit odd when `nodefaults` gets passed to as an action arg, but it is internally consistent with the existing behaviour where the provided env file is the highest authority, so it seems reasonable to me. I'm content with this and feel this is ready for review.

liamhuber:
@conda-incubator/setup-miniconda tests are passing, just waiting on a review.

dist/setup/index.js
```diff
@@ -47284,8 +47284,19 @@ function applyCondaConfiguration(inputs, options) {
         // LIFO: reverse order to preserve higher priority as listed in the option
         // .slice ensures working against a copy
         for (const channel of channels.slice().reverse()) {
-            core.info(`Adding channel '${channel}'`);
-            yield condaCommand(["config", "--add", "channels", channel], options);
+            core.info("Removing channel defaults");
+            if (channel === "nodefaults") {
```
jaimergp, line 47288:
```suggestion
            if (channel === "nodefaults") {
                core.info("Removing channel defaults");
```

src/conda.ts
```diff
@@ -126,8 +126,22 @@ export async function applyCondaConfiguration(
   // LIFO: reverse order to preserve higher priority as listed in the option
   // .slice ensures working against a copy
   for (const channel of channels.slice().reverse()) {
-    core.info(`Adding channel '${channel}'`);
-    await condaCommand(["config", "--add", "channels", channel], options);
+    core.info("Removing channel defaults");
+    if (channel === "nodefaults") {
```
jaimergp, line 130:
```suggestion
    if (channel === "nodefaults") {
      core.info("Removing channel defaults");
```

jaimergp (commented):
Just a tiny comment with the logging statements. LGTM otherwise.

jezdez:
I don't agree we should further the "nodefaults" workaround in setup-miniconda, it's a bad API since it's not an actual channel and tends to increase the risk of a messy channel management

jezdez:
I think it might be enough to simply offer "override_channels" as a config option, which requires specifying a channel but doesn't hard-code the "nodefaults" workaround.

jezdez (changes requested):
A more explicit and future-proof solution is relying on the `--override-channels` CLI option since `nodefaults` is a pretty bad workaround within conda. That way, running `conda config --remove channels defaults` isn't needed here.

liamhuber:
> Just a tiny comment with the logging statements. LGTM otherwise.

@jaimergp, nice catch! Committed 🚀 

> I don't agree we should further the "nodefaults" workaround in setup-miniconda, it's a bad API since it's not an actual channel and tends to increase the risk of a messy channel management...

@jezdez, the current situation is one "bad" API and a second API that _silently_ deviates from the first API in an unproductive way (i.e. a channel named "nodefaults" is literally added to the channels list). Regardless of whether the `conda` API is good or bad, I would strongly recommend `conda-incubator/setup-miniconda` favour consistency over goodness where the two APIs use the same terminology. 

I definitely like your suggested `--override-channels` solution, but I see it as a "yes, and" approach, and I'm afraid you'll need to identify a different volunteer for its implementation.

liamhuber:
Nice, even better, if https://github.com/conda/conda/pull/14227 is merged I think we can even simply close this PR?

jaimergp:
Yep, timing will work this time. The only other workaround right now is to explicitly pass a `condarc` file correctly populated to the `condarc-file` input.

liamhuber:
> Yep, timing will work this time. The only other workaround right now is to explicitly pass a `condarc` file correctly populated to the `condarc-file` input.

Haha, yep, did exactly that downstream a couple weeks ago! https://github.com/pyiron/actions/pull/129

oursland:
Builds are breaking in CI as the package `defaults/noarch::conda-libmamba-solver-23.12.0` is being selected, which is incompatible.

liamhuber:
@oursland I'm afraid I don't have enough context to understand your statement. The only thing breaking here was the linter test -- which failed because I never rebuilt the committed ts suggestions into the js.

I did that now, but I also see that https://github.com/conda/conda/pull/14227 got merged last week. @jaimergp my understanding is that now we just need to wait for the next release of conda, and this PR becomes completely unnecessary, yeah? If that's the case I think we can just close this PR.

oursland:
GitHub Actions should not be considered like a normal software release as this is a piece of core CI infrastructure and is not something that can always be deferred until the next release of an upstream package, which will then be packaged into a release, and eventually distribution.

The current behavior of `setup-miniconda` still breaks GitHub actions.  During update, it pulls in an incompatible version of `libmamba-solver` from the `defaults` channel.  This has been breaking CI for FreeCAD among others.

liamhuber:
> GitHub Actions should not be considered like a normal software release as this is a piece of core CI infrastructure and is not something that can always be deferred until the next release of an upstream package, which will then be packaged into a release, and eventually distribution.

I don't think I feel as strongly about it, but I also don't disagree. My opinion doesn't matter though, I'm just a rando who patched what was bothering me. If you want to get this in ASAP you'll need to convince the two reviewers.

If you want something immediately that avoids erroneously including the defaults, you can check out https://github.com/pyiron/actions/blob/main/cached-miniforge/action.yml, a wrapper for `setup-miniconda` that patches over the defaults issue and solves an edge case in caching (https://github.com/conda-incubator/setup-miniconda/issues/361)

> The current behavior of `setup-miniconda` still breaks GitHub actions. During update, it pulls in an incompatible version of `libmamba-solver` from the `defaults` channel. This has been breaking CI for FreeCAD among others.

I'm afraid I'm a little lost again. Do you mean the main branch of `setup-miniconda`, or do you mean that you tried out this branch and something I've done here broke the `libmamba-solver` versioning? If it's the main branch, raising a separate issue and referencing this PR might be the way to go; if it's something I've done here I'll need some assistance tracking down the problem -- at least a MWE for the breakage -- as I don't usually do any work in javascript.

oursland:
> I'm afraid I'm a little lost again. Do you mean the main branch of setup-miniconda, or do you mean that you tried out this branch and something I've done here broke the libmamba-solver versioning?

The main branch is broken.  This PR fixes the issue.  Waiting what could be months for to fix CI is not a reasonable strategy.

Your fix through caching is unfortunately not feasible for our application.  For FreeCAD and others, we already use 100% of our GitHub provided caching storage budget for various compiler caches.

liamhuber:
@oursland, the action I linked also removes the `defaults` channel; if this PR fixes your problem, it's likely that action will provide immediate relief as well -- the caching issue is an additional, unrelated issue it fixes. It is not a pure-wrapper though -- it will also try to pip install your local package by default -- but you should have enough flexibility in the inputs to get it to work as a pure wrapper. If you need something to solve the `defaults` problem _immediately_, I recommend trying that.

jakirkham:
AIUI Conda will go through a depreciation cycle before removal. It will probably be a bit before installers come prebaked with the newer Conda

Not to mention some users may pin an installer version as part of this workflow

Given this, think this change still is desirable. What do we need to do to get it in?

jaimergp:
Thanks @jakirkham, yes, when I posted this comment I had gone for the straight removal disregarding the deprecation cycle, which was brought in the code review.

This change is still needed as a result, but I think @jezdez doesn't want the `nodefaults` hack in more places, so we'll have to come up with a different solution.

My ideas:

- Warn about `channels` implicitly adding `defaults`
- Add a new setting to remove `defaults` if not in `channels` explicitly. Something like `ignore_defaults_in_channels`.

Since this is a new approach, I'll open a new PR. Thanks all for the fruitful discussion!

oursland:
> Add a new setting to remove defaults if not in channels explicitly. Something like ignore_defaults_in_channels.

This is the expected behavior.  If I provide a list of channels, then I expect that those channels and only those channels are being used.  The inclusion of `defaults` implicitly in the channels list was a surprise when it caused breakage in our builds.  I consider this inclusion to be a bug.

I personally do not believe an additional flag to remove `defaults` should be necessary as I consider it's implicit inclusion to be a bug that should be resolved.

jaimergp:
Believe it or not, it was intentionally designed that way. Not an oversight. You are "adding" channels to the list, which is prepopulated with `defaults`. I agree it is not great UX, so that's why `conda` is deprecating that using the standard deprecation window of three months.

If anything, the bug is here in setup-miniconda because we are using `conda config --add` to populate the channels, and it wasn't clear in the documentation that it happened that way. The option would have been better called `add_channels`. Since we can't guarantee that changing the default will break users workflows, the conservative way forward is adding a flag that people can opt-in to, and eventually turn it on by default. If you'd like to advocate for a more aggressive change, [let's discuss that in the new PR.](https://github.com/conda-incubator/setup-miniconda/pull/367)

jakirkham:
Yep that is reasonable Jaime. Am not saying it needs to be this fix, but we do need something

How about we do the same thing we do in conda-forge? Namely `conda config --remove channels defaults`, which gives us

```
# .condarc
channels: []
```

Then we can add any channels after

liamhuber:
Regardless of whether the `nodefaults` syntax is good or not, and even in the presence of the new better PR, I would actually still suggest also merging this PR. As long as conda itself supports `nodefaults` in the channels removing defaults, I feel it would be wise to synchronize behaviour in the channels arg here. Literally adding `nodefaults` to the channels list is unexpected and undesirable.

dist/setup/index.js
```diff
@@ -47284,8 +47284,19 @@ function applyCondaConfiguration(inputs, options) {
         // LIFO: reverse order to preserve higher priority as listed in the option
         // .slice ensures working against a copy
         for (const channel of channels.slice().reverse()) {
```
jakirkham, line 47286:
```suggestion
        if (channels.length) {
            yield condaCommand(["config", "--remove", "channels", "defaults"], options);
        }
        for (const channel of channels.slice().reverse()) {
```

dist/setup/index.js
```diff
@@ -47284,8 +47284,19 @@ function applyCondaConfiguration(inputs, options) {
...
+                    yield condaCommand(["config", "--remove", "channels", "defaults"], options);
+                }
+                catch (err) {
+                    core.info("Removing defaults raised an error -- it was probably not present.");
+                }
+            }
+            else {
+                core.info(`Adding channel '${channel}'`);
+                yield condaCommand(["config", "--add", "channels", channel], options);
+            }
```
jakirkham, line 47299:
```suggestion
            core.info(`Adding channel '${channel}'`);
            yield condaCommand(["config", "--add", "channels", channel], options);
```

src/conda.ts
```diff
@@ -126,8 +126,22 @@ export async function applyCondaConfiguration(
   // LIFO: reverse order to preserve higher priority as listed in the option
   // .slice ensures working against a copy
   for (const channel of channels.slice().reverse()) {
```
jakirkham, line 128:
```suggestion
if (channels.length) {
    yield condaCommand(["config", "--remove", "channels", "defaults"], options);
}
for (const channel of channels.slice().reverse()) {
```

src/conda.ts
```diff
@@ -126,8 +126,22 @@ export async function applyCondaConfiguration(
...
+        );
+      } catch (err) {
+        core.info(
+          "Removing defaults raised an error -- it was probably not present.",
+        );
+      }
+    } else {
+      core.info(`Adding channel '${channel}'`);
+      await condaCommand(["config", "--add", "channels", channel], options);
+    }
```
jakirkham, line 144:
```suggestion
    core.info(`Adding channel '${channel}'`);
    yield condaCommand(["config", "--add", "channels", channel], options);
```

jakirkham (commented):
Thanks Liam! 🙏

Perhaps a concrete suggestion makes it clearer what I'm proposing. Please let me know what you think 🙂

dist/setup/index.js
liamhuber: Doesn't this change the behaviour so that the absence of `defaults` is the new default behaviour?

I actually disagree with @oursland that including `defaults` by default is a bug that needs to be removed -- at least here. I'm 100% in favour of getting rid of it in the default behaviour of `conda` itself, but for me it's important that `setup-miniconda` be _consistent_ with `conda` where they share terminology (like "channels"). To this end I actually don't recommend changing the default behaviour, and still never want to see the channel `nodefaults` explicitly get added to the channels list.

Ie fix problems upstream, and keep behaviour downstream consistent with upstream (maybe adding extra convenience on top of that consistency as in... can't find the link on mobile, but you know, the other PR relevant here)

jakirkham: In the case when a user specifies `channels` (like below), it would only use those `channels`

https://github.com/conda-incubator/setup-miniconda/blob/8f65dda9671afc08d384e37681333ed70e2d5306/README.md?plain=1#L349-L353

If no `channels` are specified, it would fallback to the default behavior of the selected installer/Conda (typically using `defaults`)

liamhuber: But this still differs from default `conda` behviour, yeah? Where `defaults` is included by default and setting the channel `nodefaults` removes it rather than adding a channel `nodefaults`?

jakirkham: Not sure I follow

We don't specify how the `channel` public API of this GHA is implemented. It could be specified through `conda config` commands (like now) or it could be specified by writing out a `condarc` or it could be options tacked onto each `conda create`/`conda install` command or some other way not currently being considered

Think relying on how these public APIs are implemented under-the-hood is relying on an implementation detail, which would be subject to change. There is no guarantee the implementation details would remain the same

liamhuber:
It's maybe worth quickly revisiting why this PR exists: I ran into the nasty situation that `defaults` was getting added to my channels even if I didn't ask for it, #186. When I looked at `conda`, I saw this could be fixed by adding the channel `nodefaults`, but in `setup-miniconda` this simply literally added the channel `nodefaults`!

Now the real solution is to fix this upstream in `conda`, which is exactly what @jaimergp did https://github.com/conda/conda/pull/14227, so to the extend that this PR raised the visibility of #186 high enough to motivate that change, it's a full success. Everything after that real, upstream fix to the fundamental problem is just transient window dressing.

But what window dressing should there be? IMO, unless it is very explicitly documented, `setup-miniconda` should behave like `conda` where we would naively expect it to. That means I don't recommend getting rid of `defaults` by default; extending _new_ syntax to get rid of `defaults` like in #367 is great; and where `setup-miniconda` input looks like `conda` input and behaves _almost_ like conda input, it ought to behave _exactly_ like conda input -- anything else is just super confusing.

I patched the problem downstream in my own workflows by initializing an empty `.condarc`, so I've got basically zero skin in this game beyond trying to be a helpful neighbour. My feelings won't be hurt if you close this and just run with #367 until https://github.com/conda/conda/pull/14227 eventually becomes the `conda` standard -- but for the sake of other users, it's my earnest recommendation to drive for consistency between `setup-conda` and `conda` above other considerations.

dist/setup/index.js
jakirkham: Though did miss we do say we will add `defaults` to the end of the list 🤦‍♂️ 

https://github.com/conda-incubator/setup-miniconda/blob/8f65dda9671afc08d384e37681333ed70e2d5306/README.md?plain=1#L335-L337

That said, I think this is worth breaking. It would mean using a major version bump

jakirkham:
If we are going to add this with a warning, should we instead tailor the warning to...

1. Warn users that don't have defaults explicitly listed (and are getting it automatically), that they will need to in the next Miniforge major release
2. Include this fix and warn that `nodefaults` will become standard behavior and the option will go away in the next release

Just thinking fewer deprecation cycles and more targeted at the end result that we want

liamhuber:
That sounds wise. Feel free to make a suggestion in the ts with the message worded how you want it, and I'm happy to commit it and rebuild the js

jaimergp:
Closes by #367

PR #367, opened 2024-09-28
Add 'conda-remove-defaults' setting and support 'nodefaults' as a keyword channel

jaimergp:
Supersedes https://github.com/conda-incubator/setup-miniconda/pull/364

### DIFF
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        no-implicit-channels: [true, false]
+        no-implicit-channels: ["true", "false"]
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -51,11 +51,19 @@ jobs:
           python-version: "3.12"
       - shell: bash -el {0}
         run: |
-          cat <<EOF > script.py
+          cat <<EOF | python
           import json
           from subprocess import check_output
 
-          out = check_output(["conda", "config", "--show", "channels", "--json"])
+          out = check_output(
+            [
+              "conda${{ runner.os == 'Windows' && '.exe' || ''}}",
+              "config",
+              "--show",
+              "channels",
+              "--json",
+            ]
+          )
           channels = json.loads(out)["channels"]
           input_channels = "${{ matrix.channels }}".split(",")
           remove_defaults = "${{ matrix.conda-remove-defaults }}" == "true"
@@ -70,4 +78,3 @@ jobs:
           elif "nodefaults" in input_channels or remove_defaults:
             assert "defaults" not in channels
           EOF
-          python script.py
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -24,7 +24,9 @@ jobs:
     if:
       (github.event_name == 'schedule' && github.repository ==
       'conda-incubator/setup-miniconda') || (github.event_name != 'schedule')
-    name: Ex14 (${{ matrix.os }}, ${{ matrix.conda-remove-defaults }})
+    name:
+      Ex14 (${{ matrix.os }}, remove=${{ matrix.conda-remove-defaults }},
+      channels=${{ matrix.channels }})
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -47,14 +49,20 @@ jobs:
           channels: ${{ matrix.channels }}
           python-version: "3.12"
           conda-remove-defaults: ${{ matrix.conda-remove-defaults }}
-      - shell: bash -el {0}
+      - shell: bash -elc "python {0}"
         run: |
-          channels=$(conda config --show channels --json | jq -r '.channels | join(",")')
-          echo $channels
-          if [[ "${{ matrix.conda-remove-defaults }}" == "true" ]]; then
-            if [[ "$channels" != "conda-forge" ]]; then
-              exit 1
-            fi
-          elif [[ "$channels" != "conda-forge,defaults" ]]; then
-              exit 1
-          fi
+          from conda.base.context import context
+
+          channels = context.channels
+          input_channels = "${{ matrix.channels }}"
+          remove_defaults = "${{ matrix.conda-remove-defaults }}" == "true"
+
+          print("Input channels:", input_channels)
+          print("Input remove-defaults:", remove_defaults)
+          print("Computed channels":, channels)
+
+          assert "nodefaults" not in channels
+          if "defaults" in input_channels:
+            assert "defaults" in channels
+          elif "nodefaults" in input_channels or remove_defaults:
+            assert "defaults" not in channels
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -47,8 +47,8 @@ jobs:
         continue-on-error: true
         with:
           channels: ${{ matrix.channels }}
-          python-version: "3.12"
           conda-remove-defaults: ${{ matrix.conda-remove-defaults }}
+          activate-environment: ""
       - shell: bash -elc "python {0}"
         run: |
           from conda.base.context import context
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -59,7 +59,7 @@ jobs:
 
           print("Input channels:", input_channels)
           print("Input remove-defaults:", remove_defaults)
-          print("Computed channels":, channels)
+          print("Computed channels:", channels)
 
           assert "nodefaults" not in channels
           if "defaults" in input_channels:
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -44,6 +44,7 @@ jobs:
       - shell: bash -el {0}
         run: |
           channels=$(conda config --show channels --json | jq -r '.channels | join(",")')
+          echo $channels
           if [[ "${{ matrix.no-implicit-channels }}" == "true" ]]; then
             if [[ "$channels" != "conda-forge" ]]; then
               exit 1
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -51,7 +51,7 @@ jobs:
           python-version: "3.12"
       - shell: bash -el {0}
         run: |
-          cat <<EOF | python
+          cat <<EOF > script.py
           import json
           from subprocess import check_output
 
@@ -70,3 +70,4 @@ jobs:
           elif "nodefaults" in input_channels or remove_defaults:
             assert "defaults" not in channels
           EOF
+          python script.py
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -53,11 +53,12 @@ jobs:
         run: |
           cat <<EOF | python
           import json
+          import os
           from subprocess import check_output
 
           out = check_output(
             [
-              "conda${{ runner.os == 'Windows' && '.exe' || ''}}",
+              os.environ["CONDA_EXE"],
               "config",
               "--show",
               "channels",
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -56,7 +56,7 @@ jobs:
 
           out = check_output(["conda", "config", "--show", "channels", "--json"])
           channels = json.loads(out)["channels"]
-          input_channels = "${{ matrix.channels }}"
+          input_channels = "${{ matrix.channels }}".split(",")
           remove_defaults = "${{ matrix.conda-remove-defaults }}" == "true"
 
           print("Input channels:", input_channels)
--- a/.github/workflows/example-14.yml
+++ b/.github/workflows/example-14.yml
@@ -48,12 +48,14 @@ jobs:
         with:
           channels: ${{ matrix.channels }}
           conda-remove-defaults: ${{ matrix.conda-remove-defaults }}
-          activate-environment: ""
+          python-version: "3.12"
       - shell: bash -elc "python {0}"
         run: |
-          from conda.base.context import context
+          import json
+          from subprocess import check_output
 
-          channels = context.channels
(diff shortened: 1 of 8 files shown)


jaimergp:
Errors in the mamba examples are unrelated and caused by the mamba 2.0 release not having `mamba.bat` under `condabin/` anymore. Adding it at https://github.com/conda-forge/mamba-feedstock/pull/251

jakirkham:
Thanks Jaime! 🙏

Am actually wondering whether we need a setting for this or if we should just treat an explicit channel as just that what a user intended

Think it is easier and clearer for users to add `defaults` if they want it in the channel list

jaimergp:
Yes, in the long term, that's where I want to get at. My current concern if that if we change that behaviour overnight without warning first, then we might break a few CI pipelines that were inadvertently relying on defaults as a channel (because [strict priority is not default either](https://github.com/conda/conda/issues/14274)). 

After all this action is relied on by...  [16k+ workflows](https://github.com/search?q=%22uses%3A+conda-incubator%2Fsetup-miniconda%22+path%3A.github%2Fworkflows%2F&ref=opensearch&type=code). Of those, [~6k set `channels`](https://github.com/search?q=%22uses%3A+conda-incubator%2Fsetup-miniconda%22+path%3A.github%2Fworkflows%2F+%22channels%3A%22&ref=opensearch&type=code). Of those, [~3.7k workflows specify `defaults` explicitly](https://github.com/search?q=%22uses%3A+conda-incubator%2Fsetup-miniconda%22+path%3A.github%2Fworkflows%2F+%2Fchannels%3A.*defaults.*%2F&ref=opensearch&type=code).

jaimergp:
Fix for example 6 available at #368. Once that's merged we can continue here.

dbast (approved):

action.yml
```diff
@@ -193,6 +193,12 @@ inputs:
       for more information."
     required: false
     default: ""
+  no-implicit-channels:
```
jezdez, line 196:
```suggestion
  conda-remove-defaults:
```

README.md
```diff
@@ -598,6 +603,35 @@ jobs:
           python -c "import platform; assert platform.machine() == 'arm64', platform.machine()"
 ```
 
+### Example 14: No implicit channels
```
jezdez, line 606:
```suggestion
### Example 14: Remove `defaults` channel
```

README.md
```diff
@@ -598,6 +603,35 @@ jobs:
...
+        os: ["ubuntu-latest"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./
+        id: setup-miniconda
+        continue-on-error: true
+        with:
+          miniforge-version: latest
+          channels: conda-forge
+          no-implicit-channels: "true"
```
jezdez, line 628:
```suggestion
          conda-remove-defaults: "true"
```

.github/workflows/example-14.yml
```diff
@@ -0,0 +1,54 @@
...
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./
+        id: setup-miniconda
+        continue-on-error: true
+        with:
+          channels: conda-forge
+          python-version: "3.12"
+          no-implicit-channels: ${{ matrix.no-implicit-channels }}
```
jezdez, line 43:
```suggestion
          conda-remove-defaults: ${{ matrix.conda-remove-defaults }}
```

.github/workflows/example-14.yml
```diff
@@ -0,0 +1,54 @@
...
+    if:
+      (github.event_name == 'schedule' && github.repository ==
+      'conda-incubator/setup-miniconda') || (github.event_name != 'schedule')
+    name: Ex14 (${{ matrix.os }}, ${{ matrix.no-implicit-channels }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        no-implicit-channels: ["true", "false"]
```
jezdez, line 33:
```suggestion
        conda-remove-defaults: ["true", "false"]
```

.github/workflows/example-14.yml
```diff
@@ -0,0 +1,54 @@
...
+    ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  example-14:
+    # prevent cronjobs from running on forks
+    if:
+      (github.event_name == 'schedule' && github.repository ==
+      'conda-incubator/setup-miniconda') || (github.event_name != 'schedule')
+    name: Ex14 (${{ matrix.os }}, ${{ matrix.no-implicit-channels }})
```
jezdez, line 27:
```suggestion
    name: Ex14 (${{ matrix.os }}, ${{ matrix.conda-remove-defaults }})
```

.github/workflows/example-14.yml
```diff
@@ -0,0 +1,54 @@
...
+        continue-on-error: true
+        with:
+          channels: conda-forge
+          python-version: "3.12"
+          no-implicit-channels: ${{ matrix.no-implicit-channels }}
+      - shell: bash -el {0}
+        run: |
+          channels=$(conda config --show channels --json | jq -r '.channels | join(",")')
+          echo $channels
+          if [[ "${{ matrix.no-implicit-channels }}" == "true" ]]; then
```
jezdez, line 48:
```suggestion
          if [[ "${{ matrix.conda-remove-defaults }}" == "true" ]]; then
```

README.md
```diff
@@ -59,6 +59,7 @@ possibility of automatically activating the `test` environment on all shells.
 | [Caching packages](#caching-packages)                              | [![Caching Example Status][caching-badge]][caching]             |
 | [Caching environments](#caching-environments)                      | [![Caching Env Example Status][caching-env-badge]][caching-env] |
 | [Apple Silicon](#example-13-apple-silicon)                         | [![Apple Silicon][ex13-badge]][ex13]                            |
+| [No implicit channels](#example-14-no-implicit-channels)           | [![No implicit channels][ex14-badge]][ex14]                     |
```
jezdez, line 62:
```suggestion
| [Remove defaults](#example-14-conda-remove-defaults)           | [![Remove defaults][ex14-badge]][ex14]                     |
```

src/input.ts
```diff
@@ -92,6 +92,7 @@ export async function parseInputs(): Promise<types.IActionInputs> {
     minicondaVersion: core.getInput("miniconda-version"),
     miniforgeVariant: core.getInput("miniforge-variant"),
     miniforgeVersion: core.getInput("miniforge-version"),
+    noImplicitChannels: core.getInput("no-implicit-channels"),
```
jezdez, line 95:
```suggestion
    condaRemoveDefaults: core.getInput("conda-remove-defaults"),
```

src/types.ts
```diff
@@ -84,6 +84,7 @@ export interface IActionInputs {
   readonly minicondaVersion: string;
   readonly miniforgeVariant: string;
   readonly miniforgeVersion: string;
+  readonly noImplicitChannels: string;
```
jezdez, line 87:
```suggestion
  readonly condaRemoveDefaults: string;
```

src/conda.ts
```diff
@@ -166,6 +166,20 @@ export async function applyCondaConfiguration(
     await condaCommand(["config", "--add", "channels", channel], options);
   }
 
+  if (inputs.noImplicitChannels === "true" && !channels.includes("defaults")) {
```
jezdez, line 169:
```suggestion
  if (inputs.condaRemoveDefaults === "true" && !channels.includes("defaults")) {
```

dist/setup/index.js
```diff
@@ -47355,6 +47355,15 @@ function applyCondaConfiguration(inputs, options) {
             core.info(`Adding channel '${channel}'`);
             yield condaCommand(["config", "--add", "channels", channel], options);
         }
+        if (inputs.noImplicitChannels === "true" && !channels.includes("defaults")) {
```
jezdez, line 47358:
```suggestion
        if (inputs.condaRemoveDefaults === "true" && !channels.includes("defaults")) {
```

dist/setup/index.js
```diff
@@ -48188,6 +48197,7 @@ function parseInputs() {
             minicondaVersion: core.getInput("miniconda-version"),
             miniforgeVariant: core.getInput("miniforge-variant"),
             miniforgeVersion: core.getInput("miniforge-version"),
+            noImplicitChannels: core.getInput("no-implicit-channels"),
```
jezdez, line 48200:
```suggestion
            condaRemoveDefaults: core.getInput("conda-remove-defaults"),
```

action.yml
```diff
@@ -193,6 +193,12 @@ inputs:
...
+    description:
+      "Postprocess channels list to remove channels that were not added
+      explicitly in the 'channels' setting. Will default to 'true' in 2025.03."
```
jezdez, line 199:
```suggestion
      "Postprocess channels list to remove 'defaults' if it was added
      implicitely to the 'channels' setting. Will default to 'true' to
      aligned with related change in conda 25.3.0."
```

jezdez (commented):
I'd suggest to be.. more explicit with the option name, since all it does right now is removing "defaults", nothing else. Also prefix the option with `conda` since this only applies to conda, not mamba.

.github/workflows/example-14.yml
```diff
@@ -0,0 +1,54 @@
+name: "Example 14: Remove implicitly added channels"
```
jaimergp, line 1:
```suggestion
name: "Example 14: Remove defaults channel"
```

.github/workflows/example-14.yml
```diff
@@ -0,0 +1,60 @@
...
+    if:
+      (github.event_name == 'schedule' && github.repository ==
+      'conda-incubator/setup-miniconda') || (github.event_name != 'schedule')
+    name: Ex14 (${{ matrix.os }}, ${{ matrix.conda-remove-defaults }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        conda-remove-defaults: ["true", "false"]
```
jakirkham, line 33: How about something like

```suggestion
        explicitly-listed-channels-only: ["true", "false"]
```

Admittedly verby, but I think we can write this in a way that gets closer to the long-term intended behavior

jaimergp: See https://github.com/conda-incubator/setup-miniconda/pull/367#pullrequestreview-2347356646. I don't particularly care about the name itself, given it'll be probably be obsolete by March next year.

jakirkham:
> Yes, in the long term, that's where I want to get at. My current concern if that if we change that behaviour overnight without warning first, then we might break a few CI pipelines that were inadvertently relying on defaults as a channel (because [strict priority is not default either](https://github.com/conda/conda/issues/14274)).
> 
> After all this action is relied on by... [16k+ workflows](https://github.com/search?q=%22uses%3A+conda-incubator%2Fsetup-miniconda%22+path%3A.github%2Fworkflows%2F&ref=opensearch&type=code). Of those, [~6k set `channels`](https://github.com/search?q=%22uses%3A+conda-incubator%2Fsetup-miniconda%22+path%3A.github%2Fworkflows%2F+%22channels%3A%22&ref=opensearch&type=code). Of those, [~3.7k workflows specify `defaults` explicitly](https://github.com/search?q=%22uses%3A+conda-incubator%2Fsetup-miniconda%22+path%3A.github%2Fworkflows%2F+%2Fchannels%3A.*defaults.*%2F&ref=opensearch&type=code).

AIUI the path proposed here is...

1. Add a new flag and document
2. Recommend users start setting (maybe with a warning)
3. Change the default in a major version
4. Deprecate the flag
5. Remove the flag (in the next major version)

It is certainly a valid way to do it. Am just concerned by the length of time that will take and the effort it puts us and users. Can we cutdown this process a bit?

Admittedly there is the other extreme, change behavior in a major version with documentation. This is not an unheard of approach though we need not do that either

Think there is a reasonable middle path. We already are deprecating `nodefaults`. Why not leverage it for one last release then do a major version bump that drops it everywhere?

If not that, can we thin this process a bit? How can we get this to one deprecation cycle?

jaimergp:
In a way we are tied to the conda deprecation cycle, where this behavior won't be default til March 2025. We might need to keep the flag around for users of older versions.

> Think there is a reasonable middle path. We already are deprecating nodefaults. Why not leverage it for one last release then do a major version bump that drops it everywhere?

This looks reasonable to me, but the major version bump should happen in March to align with conda. Is that too long?

jaimergp:
This should be ready now:

- [Example warning for nodefaults](https://github.com/conda-incubator/setup-miniconda/actions/runs/11178230690/job/31075405441?pr=367#step:3:31)
- [Example info for conda-remove-defaults](https://github.com/conda-incubator/setup-miniconda/actions/runs/11178230690/job/31075404577?pr=367#step:3:33)
- [Example warning for implicit defaults](https://github.com/conda-incubator/setup-miniconda/actions/runs/11178230690/job/31075406250?pr=367#step:3:33)

jezdez:
I strongly disagree that `nodefaults` should be elevated to a first class feature in the action. I'm lukewarm even acknowledging it with a warning, but I guess that's fine. My main concern is that this is feature creep for a very short timeframe.

jaimergp:
Yea, I don't like it either, but there are [~200 workflows out there using it](https://github.com/search?q=%22uses%3A+conda-incubator%2Fsetup-miniconda%22+path%3A.github%2Fworkflows%2F+%2Fchannels%3A.*nodefaults.*%2F&ref=opensearch&type=code). We need to warn them about the new option, enable the behavior, and then turn it into an error. 

It's not even documented in the README, while `conda-remove-defaults` is. I think this is the best compromise we can afford for now.

.github/workflows/example-14.yml
jezdez: I think the long-term future for this feature is only to migrate conda users to the same pattern as mamba and pixi follow, setting a default channel in the distribution/bootstrapper layer and not hard-coding it in the codebase. That's why I think the name of the parameter for setup-miniconda is okay to focus on conda only, and not required to be named generically.

jezdez (approved):

PR #398, opened 2025-05-06
Check all `.condarc` files when removing `defaults`

marcoesters:
Miniconda recently added `defaults` to the `.condarc` file in the installation root directory. This causes the feature to remove `defaults` from the channels list to fail (see [Example 14](https://github.com/conda-incubator/setup-miniconda/actions/runs/14866976837/job/41746085322?pr=397)) because `setup-miniconda` assumes that the `defaults` channel is only inside `~/.condarc`.

This PR adds a more robust check to remove `defaults` from all `.condarc` files.


marcoesters:
> You'll need to build the distribution of the typescript library following the process in `CONTRIBUTING.md` so it shows up in `dist/`

I missed them when adding them to staging because I use `git add -p`.

jezdez (approved):

PR #467, opened 2026-03-17
Update delete code to move the extracted pkgs instead of removing the 1 by 1

### LINKED ISSUE #277: Post Run conda-incubator/setup-miniconda@v2 step ~150 times slower on windows-latest than on ubuntu-latest
Setup on ubuntu-latest and windows-latest:
```bash
      - uses: conda-incubator/setup-miniconda@v2
        with:
          miniforge-variant: Mambaforge
          miniforge-version: latest
          use-mamba: true
          activate-environment: modin
          environment-file: environment-dev.yml
          python-version: 3.8
          channel-priority: strict
          # we set use-only-tar-bz2 to false in order for conda to properly find new packages to be installed
          # for more info see https://github.com/conda-incubator/setup-miniconda/issues/264
          use-only-tar-bz2: false
```

Exact times: 8m54s vs 3s.

Is this behavior expected? Can something be done about it?

Thanks in advance.

### LINKED ISSUE #280: Cleanup on windows taking extremely long time
We are seeing 6-7 minute times for the cleanup job on Windows:

Screenshot 2023-03-06 at 17 40 03

 It's possible to watch every individual package deletion slowly go by one at a time.  By contrast Post Prepare on ubuntu and osx takes ~2s and ~30s, respectively. Is this time for cleanup on windows expected or unavoidable? If not is there anything we can do on our end?

### LINKED ISSUE #380: Slow cleanup / post run on windows ci since recently
Hello dearest maintainers of this very useful action!

Not sure if it's a me issue, or something else, but since recently I noticed that `post run` takes a long time (multiple hours). It looks like it has a hard time deleting some packages and sometimes failing to delete them and moving them to temp. I noticed on appveyor (with manual conda setup) that for a similar cleanup, there were suddenly a lot of permission issues, maybe that's related? There has been a new Miniforge release just a few days back ([Release 24.11.0-0](https://github.com/conda-forge/miniforge/releases/tag/24.11.0-0)) could be coincidence...

Image

goanpeca:
This fixes the slow run times of post-run on windows.

Fixes #277
Fixes #280
Fixes #380

### DIFF
--- a/src/delete.ts
+++ b/src/delete.ts
@@ -1,5 +1,4 @@
 import * as fs from "fs";
-import * as os from "os";
 import * as path from "path";
 
 import * as core from "@actions/core";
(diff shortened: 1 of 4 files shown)


src/delete.ts
```diff
@@ -8,38 +7,83 @@ import * as io from "@actions/io";
...
+    for (const rawPkgsDir of pkgsDirs) {
+      // Normalize to strip trailing separators so the sibling stash
+      // directory is always created next to pkgsDir, not inside it
+      const pkgsDir = path.resolve(rawPkgsDir);
+
+      if (!fs.existsSync(pkgsDir) || !fs.lstatSync(pkgsDir).isDirectory()) {
+        continue;
+      }
+
+      // Clean up stash directories left by previous runs (self-hosted runners)
```
jezdez: If it can't delete the old stash dirs, what's the plan to recover?

Wouldn't this be better suited for a separate action step outside the setup-miniconda action that would know exactly how to purge old stashes on a persistent runner? How can we test this?

jezdez: Option B seems like the right pragmatic choice -- on ephemeral runners it doesn't matter, and on self-hosted runners the cleanup at the start of the next run is good enough. The warning on failure is fine, worst case is some temporary disk usage until the next successful cleanup.

Option E (OS-native delete) could be worth exploring separately to see if it makes the whole stash approach unnecessary, but that shouldn't block this.

jezdez (approved):
Looks good, nice improvement. Option B for the self-hosted cleanup is the right default.

PR #453, opened 2026-02-26
Optimize conda configuration: write .condarc directly instead of spawning subprocesses

jezdez:
> **Note:** The breaking changes in this PR (minimum conda version bump, config key rename) warrant a major version bump of the action (v4).

### Fixes

- Replace `applyCondaConfiguration()` (which spawned ~13 sequential `conda config` subprocesses per run, each paying 2-5s of Python startup) with `writeCondaConfig()` that builds the complete `.condarc` in memory and writes it once via `fs.writeFileSync`.
- Remove redundant second `applyCondaConfiguration` call after base tool installation.
- Replace `conda config --show --json` subprocess in `isDefaultEnvironment` with direct YAML reads of condarc files.
- Remove `minicondaVersions()` which downloaded and parsed the full `repo.anaconda.com` HTML index page for version validation. Now attempts the download directly with a clear error on failure.
- Parallelize Windows `takeown` calls with `Promise.all`.

### Breaking changes

- Minimum `miniconda-version` bumped from 4.6.0 to 25.5.0 (required for `auto_activate` config key support).
- Writes `auto_activate` (canonical name since conda 25.5.0) instead of the deprecated `auto_activate_base` (removal scheduled for conda 26.3).

### Stability

- Existing `~/.condarc` written by the installer is now read as base config before applying action inputs, preserving installer-embedded settings (e.g., Miniforge's `channels: [conda-forge]`).
- `removeDefaults` now also strips `defaults` from prefix-level condarc files, matching the old behavior.
- Warns on unrecognized condarc keys to catch typos.
- Handles all YAML 1.1 boolean literals (`on`/`off`/`y`/`n`/`1`/`0`) in boolean config values.

### Tasks and Maintenance

- Remove `get-hrefs` and `normalize-url` dependencies.
- Remove `src/typings.d.ts` (only contained `get-hrefs` type declarations).

PR #488, opened 2026-04-15
Replace isDefaultEnvironment subprocess with local YAML reads

jezdez:
### Summary

- Replace `conda config --show --json` subprocess in `isDefaultEnvironment` with direct reads of condarc files using `js-yaml`.
- Checks user-level `~/.condarc` and prefix-level `.condarc`/`condarc` files, matching the locations conda itself consults.
- Function is now synchronous since it only does local file I/O.

### Rationale

Each `conda` subprocess incurs Python/conda startup overhead (2-5s). This function is called during `condaInit` and doesn't need the full conda runtime -- it only needs to check if `default_activation_env` is set in the config.

This is a non-breaking change extracted from #453.

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -33928,1963 +33928,6 @@ function parseInputs() {
     });
 }
 
-;// CONCATENATED MODULE: external "stream"
-const external_stream_namespaceObject = __WEBPACK_EXTERNAL_createRequire(import.meta.url)("stream");
-;// CONCATENATED MODULE: ./src/utils.ts
-/**
- * @module utils
- * Low-level helpers for running shell commands, building conda package
- * specs, and working with package directories.
- *
- * @category Core
- */
-var utils_awaiter = (undefined && undefined.__awaiter) || function (thisArg, _arguments, P, generator) {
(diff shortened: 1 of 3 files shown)


travishathaway:
@jezdez,

I'm curious why there's so many changes to the `dist/setup/index.js` file. Is this something that just gets automatically updated like when tools like `conda smithy` run on conda-forge repositories?

I'd appreciate it if you could help me learn a little more about this file.

dbast:
@travishathaway yes, those files are generated artifacts.
- The action source lives in ./src as TypeScript, with dependencies pinned in package-lock.json.
- During development/release, we run npm run build (using @vercel/ncc), which bundles source + dependencies into self-contained JS files under ./dist.
- GitHub Actions runs the bundled files from action.yml (dist/setup/index.js and dist/delete/index.js), so no need for npm install at runtime.
- That improves startup speed and reproducibility, and reduces supply-chain exposure during action execution.
- Because of this, PRs that change action code must also commit the regenerated ./dist files.
- CI checks this by rebuilding and failing if ./dist is out of sync.`

travishathaway:
Thanks for the clarification @dbast!

travishathaway (approved):

dbast (approved):

PR #539, opened 2026-06-07
fix: look up the tool cache by tool name to fix Miniconda3 cache miss…

### LINKED ISSUE #197: Miniconda3 tool cache misses
https://github.com/conda-incubator/setup-miniconda/blob/723ab12c1d66bb893c5ac701f8e387f00b24549e/src/installer/base.ts#L49-L52

Given for instance `miniconda-version: latest` this makes https://github.com/actions/toolkit/blob/2f164000dcd42fb08287824a3bc3030dbed33687/packages/tool-cache/src/tool-cache.ts#L472-L476 to look for a path in the tool cache like `<toolcachedir>/Miniconda3-latest-Linux-x86_64.sh` instead of `<toolcachedir>/Miniconda3`, where it's actually being saved.

goanpeca:
Closes #197

## Summary

For SemVer-compatible installer versions, the installer cache is written under `tool` (via `tc.cacheFile(..., tool, version, ...)`) but was looked up by the installer **filename** (`tc.find(installerName, ...)`). Whenever a caller sets `options.tool` -- which the built-in Miniconda/Miniforge downloaders always do -- the lookup key did not match the write key, so SemVer-compatible versions missed the cache and the installer was re-downloaded on every run.

Non-SemVer labels such as `latest` are intentionally passed through unchanged and delegated to `@actions/tool-cache` semantics. `latest` remains a moving pointer and is not expected to restore from the tool cache.

## Change

`src/installer/base.ts`: look up the cache by `tool` (the same key `tc.cacheFile` writes under), so built-in installers with SemVer-compatible versions hit the cache and skip the redundant re-download.

## Tests

`src/__tests__/installer/base.test.ts`:

- New test: `tc.find` is called with the tool name (the cache key), not the filename (#197).
- New test: `latest` is passed through unchanged and delegated to `@actions/tool-cache` restore semantics.
- Updated the two arch tests to expect the tool name in the `tc.find` call.

## Verification

- `npm test` -> all pass (416)
- `npm run check` -> pass (prettier + eslint)
- `npm run build` -> pass; `dist/setup/index.js` regenerated and committed

## Note

The same one-line change is also bundled in PR #536 (where it was a "bonus" fix). If this focused PR lands first, the cache change should be dropped from #536 to avoid a merge conflict.

### DIFF
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -39032,7 +39032,11 @@ function ensureLocalInstaller(options) {
             info(`Checking for cached ${tool}@${version}...`);
             // tc.find returns the name of the directory in which
             // the cached file is located.
-            const cacheDirectoryPath = find(installerName, version, ...(options.arch ? [options.arch] : []));
+            // Look up the cache by the same key it is written under (`tool`), not the
+            // installer filename. Otherwise callers that set `options.tool` (the
+            // built-in Miniconda/Miniforge downloaders) never get a cache hit and
+            // re-download on every run (#197).
+            const cacheDirectoryPath = find(tool, version, ...(options.arch ? [options.arch] : []));
             if (cacheDirectoryPath !== "") {
                 info(`Found ${installerName} cache at ${cacheDirectoryPath}!`);
                 // Append the basename of the cached file to the directory
--- a/dist/setup/index.js
+++ b/dist/setup/index.js
@@ -39035,7 +39035,9 @@ function ensureLocalInstaller(options) {
             // Look up the cache by the same key it is written under (`tool`), not the
             // installer filename. Otherwise callers that set `options.tool` (the
             // built-in Miniconda/Miniforge downloaders) never get a cache hit and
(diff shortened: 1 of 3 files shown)


dbast:
@goanpeca This fixes the cache key mismatch for semver-compatible versions, since `tc.find(tool, version, arch)` now matches the `tc.cacheFile(..., tool, version, arch)` write key.

One nuance: `latest` still won’t cache-hit because `@actions/tool-cache` only restores SemVer-compatible versions/ranges. That seems fine since `latest` is a moving pointer, but the PR body/tests should make that boundary explicit. Suggest updating the wording from “fixes `miniconda-version: latest`” to “fixes semver-compatible installer cache hits”, and maybe adding a small test/comment showing `latest` is intentionally passed through and delegated to tool-cache semantics.

goanpeca:
@dbast thanks for the comment. Fixed!

dbast (approved):

